PR #2, opened 2023-03-23
Start OpenTelemetry

unflxw:
Here's a test setup PR draft to test this with: https://github.com/appsignal/test-setups/pull/98

I'm making a ton of decisions implicitly, please bikeshed! Should we "auto-detect" instrumentations in this way? Should we still have `disable_default_instrumentations`, or is the dependency opt-in enough?

Commits are a bit too intermingled because I did a lot of changes in one go, sorry.

---

### [Read configuration from the environment](https://github.com/appsignal/appsignal-python/commit/95630dab8b2c8e230b169d647d04fb2fbe3e22ef)

Move things around so that everything that relates to the configuration
options lives in `config.py`. Implement configuration options for
`revision` and `disable_default_instrumentations`, which will be used
in later commits. Allow for the configuration options to be read from
the environment.

### [Start OpenTelemetry on client start](https://github.com/appsignal/appsignal-python/commit/42ca5f0fb7eb41f70ca1a3a1b5da83bd5317bd35)

When `client.start()` is called, start OpenTelemetry, pointing its
exporter at the agent, which is also started by `client.start()`.

Automatic instrumentation is only provided if the corresponding
OpenTelemetry package is installed -- that's the signal to "opt in".
This keeps our package light on dependencies.

### DIFF
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 from .agent import start_agent
 from .config import from_public_environ
+from .opentelemetry import start_opentelemetry
 
 if TYPE_CHECKING:
     from typing import Unpack
@@ -16,3 +17,4 @@ class Client:
 
     def start(self):
         start_agent(self._options)
+        start_opentelemetry(self._options)
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
 from .agent import start_agent
+from .config import from_public_environ
 
 if TYPE_CHECKING:
     from typing import Unpack
@@ -11,7 +12,7 @@ class Client:
     _options: Options
 
     def __init__(self, **options: Unpack[Options]):
-        self._options = options
+        self._options = from_public_environ() | options
 
     def start(self):
         start_agent(self._options)
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -13,7 +13,7 @@ class Client:
     _options: Options
 
     def __init__(self, **options: Unpack[Options]):
-        self._options = from_system() | from_public_environ() | options
+        self._options = from_system() | options | from_public_environ()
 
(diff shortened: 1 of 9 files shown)


src/appsignal/config.py
```diff
@@ -1,10 +1,101 @@
...
+        "_APPSIGNAL_ENVIRONMENT": config.get("environment"),
+        "_APPSIGNAL_LOG_LEVEL": config.get("log_level"),
+    } | CONSTANT_PRIVATE_ENVIRON
+
+    for var, value in private_environ.items():
+        if value is not None:
+            os.environ[var] = str(value)
+
+
+def opentelemetry_resource_attributes(config: Options):
```
tombruijn, line 94: This should also return the app path resource attribute.

unflxw: Thank you! I tried to figure this out, but couldn't, and then I forgot that it needed to be done, because brain.

A quick Google shows this is probably the way to go about this: https://stackoverflow.com/questions/13699283/how-to-get-the-callers-filename-method-name-in-python

tombruijn: Sounds good, but we can't rely on the current working dir? Like we do [in the Ruby gem](https://github.com/appsignal/appsignal-ruby/blob/2fc6ba85be1e0cabc2bb8fb26469ad47d1c60243/lib/appsignal.rb#L101).

unflxw: Ah, yes, that should do. I'll go with that instead. Thank you!

tombruijn (approved):

src/appsignal/client.py
```diff
@@ -13,7 +13,7 @@ class Client:
     _options: Options
 
     def __init__(self, **options: Unpack[Options]):
-        self._options = from_public_environ() | options
+        self._options = from_system() | from_public_environ() | options
```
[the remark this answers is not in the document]
tombruijn: Okay, let's implement it like in the other integrations for now.
Let's bump the discussion on this issue: https://github.com/appsignal/integration-guide/issues/73
When my other work is done I can take the lead on that.

Issue #3, opened 2023-03-24
First release for Python package


unflxw:
Releasing an alpha version of the package to PyPI would allow us to test the publishing process, give us confidence that `pip` selects the right package based on the combination platform tag, and it's a necessary step towards getting this to a state where customers can use it.

Note that PyPI has a "test register" that you can upload to and install from. We should probably use that for testing the release process: https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives

## To do

Feel free to split any of these into its own issue. It might also be convenient to do it in the opposite order -- release an alpha to PyPI first, using Hatch, then have that experience inform how Hatch support for Mono should be.

Keep in mind that the way we use Hatch is a bit unusual, so this might be less "implement Python/Hatch support for Mono" and more "implement custom scripts support for Mono"?

- [x] Release alpha to PyPI test environment
- [x] Test that the release installs correctly

tombruijn:
Let's not make Mono support a blocker. We can do the first couple releases manually.

tombruijn:
> Keep in mind that the way we use Hatch is a bit unusual, so this might be less "implement Python/Hatch support for Mono" and more "implement custom scripts support for Mono"?

I think mono already supports custom scripts for all things by customizing the `command` option for each step.

PR #17, opened 2023-04-21
Configure mono for package

tombruijn:
Configure mono for a custom language, as it doesn't directly support Python (yet).

Add scripts to read and update the package version. These will be called by mono, PR https://github.com/appsignal/mono/pull/52.

Part of #3

### DIFF
--- a/mono.yml
+++ b/mono.yml
@@ -1,0 +1,15 @@
+---
+language: custom
+repo: "https://github.com/appsignal/appsignal-python"
+bootstrap:
+  command: ""
+clean:
+  command: "hatch clean"
+build:
+  command: "hatch run build:all"
+publish:
+  command: "hatch publish"
+test:
+  command: "hatch run test:pytest"
+read_version: "hatch version"
+write_version: "hatch version"


mono.yml
```diff
@@ -0,0 +1,15 @@
...
+clean:
+  command: "hatch clean"
+build:
+  command: "hatch build"
+publish:
+  command: "hatch publish"
+test:
+  command: "hatch run test:pytest"
+read_version: "script/read_package_version"
+write_version: "script/write_package_version"
```
unflxw, line 15: Unless I'm missing something, I believe this is re-implementing the `hatch version` command: https://hatch.pypa.io/latest/version/

mono.yml
```diff
@@ -0,0 +1,15 @@
+---
+language: custom
+repo: "https://github.com/appsignal/appsignal-python"
+bootstrap:
+  command: ""
+clean:
+  command: "hatch clean"
+build:
+  command: "hatch build"
```
tombruijn, line 9:
```suggestion
  command: "hatch run build:all"
```

mono.yml
tombruijn: > (Instead of read_version and write_version, you may want to just have a single version custom script? Hatch's behaviour of printing it if no argument is passed and updating it if an argument is passed seems reasonable-ish to me)

I figured I make two config options so you can write specific scripts and don't have to check if the script received an argument or not.

unflxw (approved):

PR #21, opened 2023-04-26
Add `py.typed` to export types

unflxw:
Turns out the typechecker will ignore the type hints declared in dependencies unless those dependencies contain an empty `py.typed` file. So users who use our library will not be able to make use of our type hints unless we add this file.

See [this article][article] and [PEP 561][pepsi] for more information.

[article]: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/
[pepsi]: https://peps.python.org/pep-0561/#packaging-type-information

[skip changeset]


tombruijn (approved):

unflxw:
Merging -- we can always revisit later. Failure is Lintje because I rebased out a commit from the branch.

PR #23, opened 2023-04-26
Set the language integration as an env var

tombruijn:
Set the language integration version as an env var, rather than a resource attribute. We don't need to the particular use case of the resource attributes here.

Also include the package's version number this way so we know which package version the reported data is sent from.

[skip changeset]

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from .__about__ import __version__
 
 import os
 from typing import cast, get_args, TypedDict, Union, Optional, Literal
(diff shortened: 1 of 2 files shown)


src/appsignal/config.py
```diff
@@ -90,6 +87,7 @@ def set_private_environ(config: Options):
         "_APPSIGNAL_APP_ENV": config.get("environment"),
         "_APPSIGNAL_APP_NAME": config.get("name"),
         "_APPSIGNAL_HOSTNAME": config.get("hostname"),
+        "_APPSIGNAL_LANGUAGE_INTEGRATION_VERSION": f"python-{__version__}",
```
unflxw, line 90: Consider moving it to the `CONSTANT_PRIVATE_ENVIRON` instead, as it's not derived from the config?

unflxw (approved):

Issue #16, opened 2023-04-21
Implement missing configuration options


unflxw:
There are configuration options that can be set in the current Python beta [for the stand-alone agent](https://docs.appsignal.com/standalone-agent/configuration/options.html) or [via OpenTelemetry](https://docs.appsignal.com/opentelemetry/configuration/options.html) but aren't currently implemented in [the Python integration's options map](https://github.com/appsignal/appsignal-python/blob/main/src/appsignal/config.py#L43-L50). We should implement those.

PR #31, opened 2023-05-02
Add config options of the list type

tombruijn:
Parse lists from environment variables, and format them again when set as private env vars for the agent.

I left out request_headers, because we should check what the default for that should be in more detail. This is more the general structure of parsing and setting config options that the client does not directly use (right now).

[skip changeset]
Part of #16
Based on #28

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -36,6 +36,13 @@ def parse_bool(value: Optional[str]) -> Optional[bool]:
     return None
 
 
+def parse_list(value: Optional[str]) -> Optional[list[str]]:
+    if value is None:
+        return None
+
+    return value.split(",")
+
+
 def parse_disable_default_instrumentations(
     value: Optional[str],
 ) -> Optional[Union[list[DefaultInstrumentation], bool]]:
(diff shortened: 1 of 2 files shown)


luismiramirez (approved):

src/appsignal/config.py
```diff
@@ -134,12 +153,20 @@ def bool_to_env_str(value: Optional[bool]):
     return str(value).lower()
 
 
+def list_to_env_str(value: Optional[list[str]]):
+    if value is None:
+        return ""
```
unflxw, line 158: Note that, by returning `""`, this means that the private environment variable will be set to an empty string. I think we should avoid setting the corresponding private environment variable when the value is `None`.

There's already a check to avoid setting the environment variable if it is `None`: https://github.com/appsignal/appsignal-python/blob/b6228190ebbfc8e9659c1873a9146e4bf849ed36/src/appsignal/config.py#L207-L209

So I think this should return `None`:

```suggestion
    if value is None:
        return None
```

(If this change is done, it should probably also be done for `bool_to_env_str`)

unflxw (approved):

src/appsignal/config.py
tombruijn: both updated 👍

Issue #36, opened 2023-05-03
Refactor config 


tombruijn:
## To do

- [ ] Add a Config class
- [ ] That config class has an Options dict
- [ ] It stores each source (system, file/initial, env) as a separate source on the config class
- [ ] It merges the config options in the class
- [ ] It validates the config options in the class

unflxw:
Might be done as part of #19, since it will require the sources bit.

tombruijn:
Let's pick this up first and do the diagnose after

Issue #34, opened 2023-05-03
Add internal logger


tombruijn:
## To do

- [ ] Add `log` option. Possible values: `file` (default) and `stdout`
- [ ] Add integrations logger for the package. Support:
  - [ ] STDOUT logger
  - [ ] File logger (default)

PR #37, opened 2023-05-03
Add log_path config option

tombruijn:
﻿## Add log_path config option

Check if the specified path is writable or not, and if not, fall back on the system tmp dir.

It just prints the warning and doesn't log it. We don't have an internal logger yet, see issue #34. It will also need to log it eventually.

While discussing this implementation, we decided we want to refactor this so there's a Config class that handles all the function calls, so that we don't need to call `validate` manually this way in the future. See issue #36. So I'm leaving a proper validation "state" in the Config for that issue.

[skip changeset] Part of #16

## Fetch tmp dir in a cross platform way

Use the `gettempdir` method to fetch the system tmp dir. This won't work on macOS the way we want to, because it will create a new tmp dir every time and we want it to point to `/tmp`.

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -109,15 +109,19 @@ DEFAULT_CONFIG = Options(
 
 
 def log_file_path(config: Options) -> Optional[str]:
-    filename = "appsignal.log"
     path = config.get("log_path")
 
-    if path and not os.access(path, os.W_OK):
-        path = None
-        print(
-            f"appsignal: Unable to write to configured '{path}'. Please check the "
-            "permissions of the 'log_path' directory."
-        )
+    if path:
+        _, ext = os.path.splitext(path)
+        if ext:
+            path = os.path.dirname(path)
+
+        if not os.access(path, os.W_OK):
+            path = None
+            print(
+                f"appsignal: Unable to write to configured '{path}'. Please check the "
+                "permissions of the 'log_path' directory."
+            )
 
     if not path:
         path = "/tmp" if platform.system() == "Darwin" else tempfile.gettempdir()
@@ -128,8 +132,7 @@ def log_file_path(config: Options) -> Optional[str]:
             "permissions of the 'log_path' directory."
         )
         return None
-
-    return "/".join([path, filename])
+    return os.path.join(path, "appsignal.log")
 
 
 def from_system() -> Options:
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 from .__about__ import __version__
 
 import os
+import platform
+import tempfile
 from typing import cast, get_args, TypedDict, Union, Optional, Literal
 
 DefaultInstrumentation = Literal[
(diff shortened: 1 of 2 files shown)


src/appsignal/config.py
```diff
@@ -104,6 +108,39 @@ class Options(TypedDict, total=False):
...
 
+def validate(config: Options):
+    config["log_file_path"] = log_file_path(config)
+
+
+def log_file_path(config: Options) -> Optional[str]:
+    filename = "appsignal.log"
+    path = config.get("log_path")
+
+    print(f"!!!path {path}")
```
luismiramirez, line 119: Debug print?

luismiramirez (approved):

src/appsignal/config.py
```diff
@@ -104,6 +108,38 @@ class Options(TypedDict, total=False):
 )
 
 
+def validate(config: Options):
+    config["log_file_path"] = log_file_path(config)
+
+
+def log_file_path(config: Options) -> Optional[str]:
+    filename = "appsignal.log"
+    path = config.get("log_path")
```
unflxw, line 117: I recall other integrations remove `appsignal.log` from the end of the path if already present. Should we do that here as well?

src/appsignal/config.py
```diff
@@ -104,6 +108,38 @@ class Options(TypedDict, total=False):
...
+        path = "/tmp" if platform.system() == "Darwin" else tempfile.gettempdir()
+
+    if not os.access(path, os.W_OK):
+        print(
+            (
+                f"appsignal: Unable to write to '{path}'. Please check the "
+                "permissions of the 'log_path' directory."
+            )
+        )
+        return None
```
unflxw, line 138: If I recall correctly, other integrations set the `log` config options (which determines where to write logs) to `stdout` instead of `file` when the logging path is not writable. I see that `log` is not implemented yet, should this be done as part of #34?

unflxw (approved):

src/appsignal/config.py
tombruijn: That was because at some point we told people to specify it. If we never tell people that for Pythong, we never support it. But we can choose to remove it anyway, so you don't get `/path/to/log/mylog.logappsignal.log` kind of scenario, but that already wouldn't work in most cases because `/path/to/log/mylog.log` probably doesn't exist and won't be writable.

src/appsignal/config.py
```diff
@@ -104,6 +108,38 @@ class Options(TypedDict, total=False):
 )
 
 
+def validate(config: Options):
+    config["log_file_path"] = log_file_path(config)
```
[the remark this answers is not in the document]
tombruijn: I can move it to the method that sets the private env vars, like in the Node.js and Ruby packages. But modifying the config in the validate method is actually [quite common for us](https://github.com/appsignal/appsignal-ruby/blob/efc7bf5e686baad83fc977417f1ff89c16a41a86/lib/appsignal/config.rb#L356-L375) 😅

src/appsignal/config.py
tombruijn: Added a commit that removes a filename if it's present.

unflxw (approved):

PR #38, opened 2023-05-03
Implement install command

### LINKED ISSUE #18: Install command
Add an install command to make the installation easier.

## To do

- [ ] Add CLI command: `install`
- [ ] Accept one CLI argument containing the Push API key
- [ ] Add an AppSignal config file, see ~docs~ test setup

unflxw:
Implement an installation command that asks the user to input the necessary information to generate an `appsignal_config.py` file.

Closes #18. [skip changeset]

The installer is functional, but it's missing "flair" -- emojis, links to the docs, friendliness, that sort of stuff. Suggestions welcome. I could also just copy what some other integration's installer does.

### DIFF
--- a/src/appsignal/cli.py
+++ b/src/appsignal/cli.py
@@ -65,15 +65,33 @@ class InstallCommand(CLICommand):
         self._name = None
 
     def run(self) -> int:
-        print("Welcome to the AppSignal for Python installer!")
+        print("👋 Welcome to the AppSignal for Python installer!")
+        print()
+        print("Reach us at support@appsignal.com for support")
+        print("Documentation available at https://docs.appsignal.com/python")
+        print()
+
         self._input_name()
 
         if self._push_api_key is None:
             self._input_push_api_key()
 
+        print()
+
         if self._should_write_file():
             print(f"Writing the {INSTALL_FILE_NAME} configuration file...")
             self._write_file()
+            print()
+            print("✅ Done! AppSignal for Python has now been installed.")
+            print()
+            print(
+                "To start AppSignal in your application, add the following code to your"
+            )
+            print("application's entrypoint:")
+            print()
+            print("    from __appsignal__ import appsignal")
+            print("    appsignal.start()")
+            print()
         else:
             print("Nothing to do. Exiting...")
 
--- a/src/appsignal/cli.py
+++ b/src/appsignal/cli.py
@@ -1,2 +1,118 @@
+from abc import ABC, abstractmethod
+import os
+
+from typing import Optional
+from docopt import docopt
+
+from .__about__ import __version__
+
+DOC = """AppSignal for Python CLI.
+
+Usage:
+  appsignal install [--push-api-key=<key>]
+  appsignal (-h | --help)
+  appsignal --version
+
+Options:
+  -h --help             Show this screen and exit.
+  --version             Show version number and exit.
+  --push-api-key=<key>  Push API Key to use for the installation.
+"""
+
+
 def run():
-    print("Hello AppSignal!")
+    version = f"AppSignal for Python v{__version__}"
+
+    arguments = docopt(DOC, version=version)
+
+    return command_for(arguments).run()
+
+
+class CLICommand(ABC):
+    @abstractmethod
+    def run(self) -> int:
+        raise NotImplementedError
+
+
+def command_for(arguments) -> CLICommand:
+    if arguments["install"]:
+        return InstallCommand(push_api_key=arguments["--push-api-key"])
+    else:
+        raise NotImplementedError
+
+
+INSTALL_FILE_TEMPLATE = """from appsignal import Appsignal
+
+appsignal = Appsignal(
+    active=True,
+    name="{name}",
+    push_api_key="{push_api_key}",
+    environment="development",
+)
+"""
+
+INSTALL_FILE_NAME = "appsignal_config.py"
+INSTALL_FILE_PATH = f"{os.getcwd()}/{INSTALL_FILE_NAME}"
+
+
+class InstallCommand(CLICommand):
+    _push_api_key: Optional[str]
+    _name: Optional[str]
+
+    def __init__(self, push_api_key=None):
+        self._push_api_key = push_api_key
+        self._name = None
+
+    def run(self) -> int:
+        print("Welcome to the AppSignal for Python installer!")
+        self._input_name()
+
+        if self._push_api_key is None:
+            self._input_push_api_key()
+
+        if self._should_write_file():
+            print(f"Writing the {INSTALL_FILE_NAME} configuration file...")
+            self._write_file()
+        else:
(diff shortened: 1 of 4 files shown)


luismiramirez (approved):

src/appsignal/cli.py
```diff
@@ -1,2 +1,118 @@
...
+        raise NotImplementedError
+
+
+INSTALL_FILE_TEMPLATE = """from appsignal import Appsignal
+
+appsignal = Appsignal(
+    active=True,
+    name="{name}",
+    push_api_key="{push_api_key}",
+    environment="development",
```
tombruijn, line 50: Is there like a `PYTHON_ENV` or other way to dynamically determine the app env?

src/appsignal/cli.py
```diff
@@ -1,2 +1,118 @@
...
+    def __init__(self, push_api_key=None):
+        self._push_api_key = push_api_key
+        self._name = None
+
+    def run(self) -> int:
+        print("Welcome to the AppSignal for Python installer!")
+        self._input_name()
+
+        if self._push_api_key is None:
+            self._input_push_api_key()
```
tombruijn, line 71: Eventually we should validate this key's validity and exit if it's not. We can create a new issue for that and pick it up later. I don't see it as a blocker.

tombruijn:
Figured out how to test it locally: `hatch run appsignal install`

Is the package configured through the same method to expose an "appsignal" executable when the package is installed?

> The installer is functional, but it's missing "flair" -- emojis, links to the docs, friendliness, that sort of stuff. Suggestions welcome. I could also just copy what some other integration's installer does.

I do not mind the lack of flair or emojis. I don't need a circus of emoji and colorized output when running an installer.

src/appsignal/cli.py
```diff
@@ -1,2 +1,118 @@
...
+class InstallCommand(CLICommand):
+    _push_api_key: Optional[str]
+    _name: Optional[str]
+
+    def __init__(self, push_api_key=None):
+        self._push_api_key = push_api_key
+        self._name = None
+
+    def run(self) -> int:
+        print("Welcome to the AppSignal for Python installer!")
```
tombruijn, line 67:
```suggestion
        print("Welcome to the AppSignal for Python installer!\n")
        print("Reach us at support@appsignal.com for support\n")
        print("Documentation available at https://docs.appsignal.com/python")
```

src/appsignal/cli.py
unflxw: I don't think so. :( All I see online, googling for things like "python production development environment", is people implementing it themselves with some custom environment variable. Django does [something weird](https://docs.djangoproject.com/en/4.2/topics/settings/#envvar-DJANGO_SETTINGS_MODULE) to pick different configurations, which I'm not sure we can derive an environment from.

If it turns out there is a way, we should remove it from the installer-generated file and have the integration itself auto-detect it.

src/appsignal/cli.py
unflxw: Ah, yes! I may give it a try now. If it's too hard then I'll create an issue. But I don't think it will be hard because Requests is wonderful.

unflxw:
> @tombruijn Figured out how to test it locally: `hatch run appsignal install`
> Is the package configured through the same method to expose an "appsignal" executable when the package is installed?

Sorry, forgot to mention how to test it locally 🙃 And, well, today I learned! The way I've been testing it myself is by running `hatch shell`, which drops you in a Python virtualenv with your package installed, and then running `appsignal install`. I like the way you found better.

So, yes, it exposes an "appsignal" executable when the package is installed, like the Ruby gem does.

tests/test_cli.py
```diff
@@ -0,0 +1,138 @@
...
+
+    yield
+
+    assert prompt_calls == cast(MagicMock, builtins.input).mock_calls
+
+
+def mock_file_operations(mocker, file_exists=False):
+    mocker.patch("os.path.exists", return_value=file_exists)
+    mocker.patch("os.getcwd", return_value="/app")
+    mocker.patch("builtins.open")
```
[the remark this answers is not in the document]
unflxw: It does fail if I ask for `builtins.bogus`. But I don't think it guarantees that we're using it "the right way" (basically, all the ways that you could possibly interact with the mock object will return more mock objects, I think even if you were to call methods that the original doesn't have)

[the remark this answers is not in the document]
tombruijn: That's better than calling `builtins.bogus` and it being mocked to pass the test.

src/appsignal/cli.py
tombruijn: > I don't think so. :(

😔 Then we should document that people need to customize that. https://github.com/appsignal/appsignal-docs/issues/1285

> If it turns out there is a way, we should remove it from the installer-generated file and have the integration itself auto-detect it.

Yes, and we can still let users customize it then.

unflxw: Yep, it will need to be done by the user manually. Maybe instead of putting it in the config file, we should have `environment="development"` as one of the config defaults? And then users can override it with `APPSIGNAL_APP_ENV`? Or is it better to have it set here, in the installer? 🤔

tombruijn: Development is a good default. If every app has its own way of determine the env, it's up to the devs to configure AppSignal correctly 🤷‍♂️
They can do it just as well in this file with a conditional.

src/appsignal/cli.py
```diff
@@ -1,2 +1,118 @@
...
+
+appsignal = Appsignal(
+    active=True,
+    name="{name}",
+    push_api_key="{push_api_key}",
+    environment="development",
+)
+"""
+
+INSTALL_FILE_NAME = "appsignal_config.py"
```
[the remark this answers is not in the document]
unflxw: I have to look into this, because I would like to call it `appsignal.py` instead of `appsignal_config.py` as well.

But this file needs to import the `appsignal` *module*, the package, to do `from appsignal import Appsignal`. Because this file is also named `appsignal`, it shadows the module, and the file tries to import itself. 🤦 

But, relative imports, right? Well, you can only use those in a module context -- meaning you can't use them from Django's `manage.py`, as we do in the test setup.

There has to be a solution for this, I'll take a look.

src/appsignal/cli.py
unflxw: Adding it to the default config and leaving it out of the generated file then. 👍

tombruijn: I wasn't clear: we can move it to the default config and leave it out of this generated file. We have a similar default in Node.js.

src/appsignal/cli.py
unflxw: I don't quite understand that. How can the installer support those if we don't have a config yet? (I suppose if given as environment variables?)

Anyway, ETOOHARD: https://github.com/appsignal/appsignal-python/issues/42

PR #39, opened 2023-05-03
Add request_headers config option

tombruijn:
Allow users to specify an allow list of request headers. Configure the OpenTelemetry instrumentation using the env var. There is no config option on the Django or wsgi/asgi instrumentations to specify it another way, only the env var.

[skip changeset]
Part of #16


src/appsignal/client.py
```diff
@@ -18,4 +24,12 @@ def __init__(self, **options: Unpack[Options]):
     def start(self):
         if self._options.get("active"):
             start_agent(self._options)
+
+            # Configure OpenTelemetry request headers config
+            request_headers = list_to_env_str(self._options.get("request_headers"))
+            if request_headers:
+                os.environ[
+                    "OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST"
+                ] = request_headers
```
unflxw, line 33: Could this be done inside `start_opentelemetry`, or in a function called by `start_opentelemetry`? The comment explaining what it's about makes me think it could be a function name instead.

Since this config is for the Django instrumentation, I'm thinking it could be done in `add_django_instrumentation`. I'd also be happy with doing it directly in `start_opentelemetry` or in a `set_opentelemetry_environ` function.

unflxw (approved):

luismiramirez (approved):

Issue #19, opened 2023-04-21
Add diagnose command


tombruijn:
Add a CLI command for the diagnose

## To do, minimum

- [ ] Add CLI command: `diagnose`
- [ ] Print output about (check the [integrations guide](https://github.com/appsignal/integration-guide/blob/main/features/cli/diagnose.md) and other languages for the specifics):
  - [ ] Library and language
  - [ ] Host
  - [ ] Agent diagnose report
  - [ ] Configuration
  - [ ] Push API key validation
  - [ ] Paths
- [ ] Add an `--environment` option? Other languages have this.
- [ ] Add a diagnose transmitter
- [ ] Add a `--[no-]send-report` option to automatically send the report
- [ ] Add Python to the diagnose_tests repo and add it as a submodule

## To do, secondary
Create an issue for this if we want to split up the work.

- [ ] Update the front-end to render the Python diagnose report

tombruijn:
Diagnose command was merged in https://github.com/appsignal/appsignal-python/pull/89

PR #45, opened 2023-05-08
Refactor config into Config class

### LINKED ISSUE #36: Refactor config

tombruijn:
We had a bunch of different Options dict instances that we merged for multiple sources to create the config for the package.

This change merges all config into one class that keeps track of the config options, sources, and potential other state.

[skip changeset]
Needed for #19, to implement the diagnose report.
Closes #36

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -4,62 +4,10 @@ from .__about__ import __version__
 import os
 import platform
 import tempfile
-from typing import cast, get_args, TypedDict, Union, Optional, Literal
+from typing import TYPE_CHECKING, cast, get_args, TypedDict, Union, Optional, Literal
 
-DefaultInstrumentation = Literal[
-    "opentelemetry.instrumentation.celery",
-    "opentelemetry.instrumentation.django",
-    "opentelemetry.instrumentation.jinja2",
-    "opentelemetry.instrumentation.psycopg2",
-    "opentelemetry.instrumentation.redis",
-    "opentelemetry.instrumentation.requests",
-]
-
-DEFAULT_INSTRUMENTATIONS = cast(
-    list[DefaultInstrumentation], list(get_args(DefaultInstrumentation))
-)
-
-CONSTANT_PRIVATE_ENVIRON = {
-    "_APPSIGNAL_LANGUAGE_INTEGRATION_VERSION": f"python-{__version__}",
-    "_APPSIGNAL_ENABLE_OPENTELEMETRY_HTTP": "true",
-}
-
-
-def parse_bool(value: Optional[str]) -> Optional[bool]:
-    if value is None:
-        return None
-
-    val = value.lower()
-    if val == "true":
-        return True
-    if val == "false":
-        return False
-
-    return None
-
-
-def parse_list(value: Optional[str]) -> Optional[list[str]]:
-    if value is None:
-        return None
-
-    return value.split(",")
-
-
-def parse_disable_default_instrumentations(
-    value: Optional[str],
-) -> Optional[Union[list[DefaultInstrumentation], bool]]:
-    if value is None:
-        return None
-
-    if value.lower() == "true":
-        return True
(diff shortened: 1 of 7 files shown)


src/appsignal/config.py
```diff
@@ -96,100 +44,218 @@ class Options(TypedDict, total=False):
...
-    enable_host_metrics=True,
-    enable_nginx_metrics=False,
-    enable_statsd=False,
-    endpoint="https://push.appsignal.com",
-    files_world_accessible=True,
-    send_environment_metadata=True,
-    send_params=True,
-    send_session_data=True,
-)
+class Config:
```
unflxw, line 47: Consider [adding type hints](https://stackoverflow.com/a/47161395) for the instance properties:
```suggestion
class Config:
  initial_source: Optional[Options]
  system_source: Options
  environment_source: Options
  options: Options
```

src/appsignal/config.py
```diff
@@ -96,100 +44,218 @@ class Options(TypedDict, total=False):
...
+        "opentelemetry.instrumentation.requests",
+    ]
+    DEFAULT_INSTRUMENTATIONS = cast(
+        list[DefaultInstrumentation], list(get_args(DefaultInstrumentation))
+    )
 
-def log_file_path(config: Options) -> Optional[str]:
-    path = config.get("log_path")
+    def __init__(self, options: Optional[Options]):
+        self.initial_source = options
```
unflxw, line 76: If you set the `initial_source` variable to an empty Options object if it is `None`, then it will be possible to treat it like any other source (from the diagnose report, for example) without having to worry about it possibly being `None`. Also, it saves you the `if self.initial_source:` later in this method, when merging the option sources.

```suggestion
        self.initial_source = options || Options()
```

src/appsignal/config.py
```diff
@@ -96,100 +44,218 @@ class Options(TypedDict, total=False):
...
+        "opentelemetry.instrumentation.redis",
+        "opentelemetry.instrumentation.requests",
+    ]
+    DEFAULT_INSTRUMENTATIONS = cast(
+        list[DefaultInstrumentation], list(get_args(DefaultInstrumentation))
+    )
 
-def log_file_path(config: Options) -> Optional[str]:
-    path = config.get("log_path")
+    def __init__(self, options: Optional[Options]):
```
unflxw, line 75: Confusingly, the type `Optional` means that it can be `None` (like `Option` in Rust) but as you've probably noticed, it doesn't make the parameter itself is optional. For that, like in Ruby, you need to give it a default value, so that you don't need to pass `None` explicitly:
```suggestion
    def __init__(self, options: Optional[Options] = None):
```

unflxw (approved):
Looks good overall! Left some nits on code stuff. Other thoughts (none of these _need_ doing):

- On the Node.js instrumentation at least, the config object has a map of sources (`{"initial": ..., "system": ...}`) -- this is later used by the diagnose report, where the maps' key is used as the name of the source. We might want to implement it like that here as well, instead of having an instance variable for each source -- the diagnose report will need to name them. (But we can do this part when we do the diagnose report.)
- We could implement `.get` and/or the `__getitem__` magic method on the `Config` class directly for convenience, so we don't have to do `config.options[whatever]` or `config.options.get(whatever)`
- `load_from_system` and `load_from_environment` don't use their `self` argument, so they could be implemented as [class methods](https://docs.python.org/3/library/functions.html#classmethod).

tombruijn:
Thanks for the suggestions! I'm going to merge this and address them in a new PR :)

src/appsignal/config.py
unflxw: My bad! Yes, it's `or`. My brain got Ruby-ed for a second.

PR #46, opened 2023-05-08
Config refactor part 2

tombruijn:
## Set default options for Config initializer

No need to pass in `None` as an argument when no Options need to be set.

[skip changeset]

## Make `load_from_<source>` methods class methods

They don't need to know anything of the instance state currently, so let's leave it out.

## Move config sources to shared sources dictionary

Rather than assign separate fields on the Config class, store all different config sources on the same dictionary so it's easier to find back.

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -74,14 +74,17 @@ class Config:
     )
 
     def __init__(self, options: Optional[Options] = None):
-        self.initial_source = options or Options()
-        self.system_source = Config.load_from_system()
-        self.environment_source = Config.load_from_environment()
+        self.sources = dict(
+            default=self.DEFAULT_CONFIG,
+            system=Config.load_from_system(),
+            initial=options or Options(),
+            environment=Config.load_from_environment(),
+        )
         self.options = (
-            self.DEFAULT_CONFIG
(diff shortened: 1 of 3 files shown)


src/appsignal/config.py
```diff
@@ -73,19 +73,26 @@ class Config:
...
-        self.system_source = self.load_from_system()
-        self.environment_source = self.load_from_environment()
-        self.options = self.DEFAULT_CONFIG | self.system_source
-        if self.initial_source:
-            self.options = self.options | self.initial_source
-        self.options = self.options | self.environment_source
-
-    def load_from_system(self) -> Options:
+    def __init__(self, options: Optional[Options] = None):
+        self.sources = dict(
```
tombruijn, line 77: Should I extract this as a new TypeDict? so it's easier to add type hints?

unflxw: I like the idea of having a Sources typed dict, but if you run into too much type casting pain, feel free to drop it.

Either way, could you could add a type hint to this `Config` class about the `sources` attribute? Either `sources: dict[string, Options]` or with the specific typed dict if you go that way.

unflxw (approved):

PR #47, opened 2023-05-08
Add convenience method to fetch config options

tombruijn:
Add `Config().option("key")` to fetch config options more easily than `Config().options.get("key")`.

I also looked into supporting `__getitem__`, but this gives trouble trying to fetch keys from a TypedDict (the options object) when the key is a variable. It expects a string literal.

[skip changeset]

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -87,6 +87,9 @@ class Config:
             | self.sources["environment"]
         )
 
+    def option(self, option: str):
+        return self.options.get(option)
+
     @classmethod
     def load_from_system(_cls) -> Options:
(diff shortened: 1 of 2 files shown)


unflxw (approved):
On second thought, I like not implementing `__getitem__` -- meaning not throwing `KeyError` if it's not found. It's kind of a foot gun.

I only wish there was a way to type-check the key, so that we could get a type error if we try to call `config.option("this_key_does_not_exist")`. On TypeScript, you would be able to type the `option` argument as `keyof Options` instead of `str`. I don't know if there's an equivalent in MyPy.

PR #50, opened 2023-05-12
Add internal logger

tombruijn:
Allow us to log things to the appsignal.log file.
Add a new "appsignal" logger in Python and configure it to match our logging standard.
https://github.com/appsignal/integration-guide/blob/e0faa6a9e02265b375a449ec6b0f234ccc68834a/build/logging.md

Use the tmp_path helper argument in pytest to create a temporary path to store the log file in for testing.

There is no "trace" level in Python. Not sure how we can support that.

This doesn't implement the STDOUT logger. That's the next step.

[skip changeset]
Part of #34

### DIFF
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -14,7 +14,7 @@ class Client:
     _logger: Logger
     _config: Config
 
-    LOG_LEVELS = dict(error=ERROR, warning=WARNING, info=INFO, debug=DEBUG)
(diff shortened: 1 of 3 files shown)


tombruijn:
> There is no "trace" level in Python. Not sure how we can support that.

If anyone has any good ideas for the trace level, I'm all ears. I don't think we can use NOTSET for that as a level.

luismiramirez (approved):
> There is no "trace" level in Python. Not sure how we can support that.

If it's out of the language conventions I think we're good without it on this integration.

PR #56, opened 2023-05-17
Add logging for internals

### LINKED ISSUE #34: Add internal logger

tombruijn:
Track if AppSignal is starting and if the OpenTelemetry instrumentations are added.

[skip changeset]
Closes #34

### DIFF
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -26,6 +26,7 @@ class Client:
 
     def start(self):
         if self._config.option("active"):
+            self._logger.info("Starting AppSignal")
             start_agent(self._config)
             start_opentelemetry(self._config)
 
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import os
 from typing import Dict
+import logging
 
 from .config import Config, list_to_env_str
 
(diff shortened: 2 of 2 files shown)


luismiramirez (approved):

conftest.py
```diff
@@ -10,3 +11,7 @@ def reset_environment_between_tests():
 
     os.environ.clear()
     os.environ.update(old_environ)
+
+    logger = logging.getLogger("appsignal")
+    for handler in logger.handlers:
+        logger.removeHandler(handler)
```
unflxw, line 17: I'd suggest moving this to a separate fixture so that it's not in the `reset_environment_between_tests` fixture. I'd also be okay with renaming the fixture to something more generic.

```suggestion

@pytest.fixture(scope="function", autouse=True)
def remove_logging_handlers_after_tests():
    yield
    
    logger = logging.getLogger("appsignal")
    for handler in logger.handlers:
        logger.removeHandler(handler)
```

unflxw (approved):
Glad you found a solution to the broken logger output issue! I took a look at it, but couldn't figure it out.

unflxw:
This print could/should also be done through the internal logger, I think: https://github.com/appsignal/appsignal-python/blob/a98c196e9964522bb5c7ead57f196228e7c9fe79/src/appsignal/agent.py#LL36C36-L36C36

PR #57, opened 2023-05-17
Listen and register agent exit code

tombruijn:
When the agent is started, track if it's started correctly. This will help debug issues and warn people when the agent can't start for whatever reason.

I've added a global `agent_active` variable which can be accessed through the `is_agent_active` helper function. It doesn't update if you import the global variable directly. Useful for the diagnose tool later on.

Listen for the exit code from the agent after starting the agent. The Python app will wait until the agent has started and exited.

This would normally cause a TimeoutError, because the agent will need to continue to run and `wait` waits for it to exit. However, turns out we still daemonize the agent in this scenario (it's not marked as standalone mode). So when it daemonizes we can assume it has started properly. If the config is invalid according to the agent, it will exit with an error before daemonizing.

This may add a small delay to the boot of the program, as long as the agent needs to start and daemonize. We can possibly do this async in the future, but I couldn't figure it out.

[skip changeset]

### DIFF
--- a/src/appsignal/agent.py
+++ b/src/appsignal/agent.py
@@ -6,9 +6,31 @@ import subprocess
 from .config import Config
 
 AGENT_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "appsignal-agent")
+agent_active = False
+
+
+def is_agent_active():
+    global agent_active
+    return agent_active
+
+
+def agent_is_active():
+    global agent_active
+    agent_active = True
 
 
 def start_agent(config: Config):
     config.set_private_environ()
-
-    subprocess.Popen([AGENT_PATH, "start", "--private"])
+    p = subprocess.Popen(
+        [AGENT_PATH, "start", "--private"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    p.wait(timeout=1)
+    returncode = p.returncode
+    if returncode == 0:
+        agent_is_active()
+    else:
+        output, _ = p.communicate()
+        out = output.decode("utf-8")
+        print(f"AppSignal agent is unable to start ({returncode}): ", out)
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,4 +1,5 @@
 from appsignal.client import Client
+from appsignal.agent import is_agent_active
 
 import os
 import re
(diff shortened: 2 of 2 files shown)


luismiramirez (approved):

src/appsignal/agent.py
```diff
@@ -6,9 +6,31 @@
...
 
 
 def start_agent(config: Config):
     config.set_private_environ()
-
-    subprocess.Popen([AGENT_PATH, "start", "--private"])
+    p = subprocess.Popen(
+        [AGENT_PATH, "start", "--private"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
```
unflxw, line 27: This potential issue was highlighted by @luismiramirez on [this Slack message](https://appsignal.slack.com/archives/C055A2H9NPJ/p1684328919356849?thread_ts=1684315640.997929&cid=C055A2H9NPJ):

> This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use Popen.communicate() when using pipes to avoid that.

The quote is in the context of using `Popen.communicate()` instead of `Popen.wait()`, to avoid a deadlock between the subprocess (waiting for the pipe to be read from so it can write more to it) and the Python application (waiting for the subprocess to end)

But what concerns me is not that we wait for it on the Python side, but that regardless, because we're not reading from that pipe, the pipe becomes full -- causing the agent subprocess to become locked trying to write from it. Is that something that could happen?

tombruijn: The agent will not run for a long time. It either exits and prints a couple lines of output, or it starts, daemonizes and starts a new process. I don't think the deadlock because of high output will ever happen.

unflxw: > or it starts, daemonizes and starts a new process

And, just to be sure, the new process wouldn't inherit this stdout/stderr file descriptor, right?

unflxw (approved):

src/appsignal/agent.py
tombruijn: I don't think so. But it also doesn't print anything after starting.

unflxw: Ah, yeah, because it always writes to a file, right? That's probably okay then.

Issue #12, opened 2023-03-27
Add demo command


unflxw:
Add a demo command to test AppSignal.

## To do

- [x] Add a CLI command: `demo`
- [x] Send an error trace (see other integrations for examples)
- [x] Send a performance trace (see other integrations for examples)
- [x] Run the demo command (or the internal code) automatically after `appsignal install` #18 
- [ ] ~Update the [integration-guide page](https://github.com/appsignal/integration-guide/blob/main/features/cli/demo.md)~

PR #59, opened 2023-05-22
Add demo command

luismiramirez:
Demo CLI command that sends two sample actions to AppSignal. One standard HTTP one and an error.

The install command now calls the demo command when finishing.

Part of: https://github.com/appsignal/appsignal-python/issues/12

[skip changeset]

### DIFF
--- a/src/appsignal/cli.py
+++ b/src/appsignal/cli.py
@@ -1,22 +1,28 @@
 from abc import ABC, abstractmethod
 import os
+import json
 
 from typing import Optional
 from docopt import docopt
 
 from .__about__ import __version__
 
+from appsignal.client import Client
+from opentelemetry import trace
+
 DOC = """AppSignal for Python CLI.
 
 Usage:
   appsignal install [--push-api-key=<key>]
+  appsignal demo [--application=<app>] [--push-api-key=<key>]
   appsignal (-h | --help)
   appsignal --version
 
 Options:
   -h --help             Show this screen and exit.
   --version             Show version number and exit.
-  --push-api-key=<key>  Push API Key to use for the installation.
+  --push-api-key=<key>  Push API Key to use for the installation and the demo.
+  --application=<app>   Application name to use for the demo.
 """
 
 
@@ -40,6 +46,11 @@ class CLICommand(ABC):
 def command_for(arguments) -> CLICommand:
     if arguments["install"]:
         return InstallCommand(push_api_key=arguments["--push-api-key"])
+    if arguments["demo"]:
+        return DemoCommand(
+            push_api_key=arguments["--push-api-key"],
+            application=arguments["--application"],
+        )
     else:
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


src/appsignal/cli.py
```diff
@@ -135,3 +150,54 @@ def _write_file(self):
             )
 
             f.write(file_contents)
+
+
+class DemoCommand(AppsignalCLICommand):
+    def __init__(self, push_api_key=None, application=None):
+        self._push_api_key = push_api_key or os.environ.get("APPSIGNAL_PUSH_API_KEY")
+        self._name = application
```
unflxw, line 158: If we listen to `APPSIGNAL_PUSH_API_KEY`, we probably also want to listen to `APPSIGNAL_APP_NAME`.

```suggestion
        self._push_api_key = push_api_key or os.environ.get("APPSIGNAL_PUSH_API_KEY")
        self._name = application or os.environ.get("APPSIGNAL_APP_NAME")
```

The demo command in the other integrations also allows passing the environment as an argument. We might want to allow that here as well. It defaults to "development" if not given, which is probably okay -- no need to ask for it via input.

src/appsignal/cli.py
```diff
@@ -135,3 +150,54 @@ def _write_file(self):
...
+        with tracer.start_as_current_span("HTTP GET") as span:
+            span.set_attribute("http.method", "GET")
+            span.set_attribute(
+                "appsignal.request.parameters",
+                json.dumps({"GET": {"id": 1}, "POST": {}}),
+            )
+            span.set_attribute(
+                "otel.instrumentation_library.name",
+                "opentelemetry.instrumentation.wsgi",
+            )
```
unflxw, line 192: Could we make it clearer to the user that these are demo spans? They should not mistake them for real performance/error samples. Maybe make the endpoint something like `GET /demo`?

src/appsignal/cli.py
```diff
@@ -135,3 +150,54 @@ def _write_file(self):
...
+
+        # Error sample
+        with tracer.start_as_current_span("HTTP GET") as span:
+            span.set_attribute("http.method", "GET")
+            span.set_attribute(
+                "appsignal.request.parameters",
+                json.dumps({"GET": {"id": 1}, "POST": {}}),
+            )
+            span.record_exception(ValueError("Something went wrong"))
```
unflxw, line 201: If the exception is not raised, it will not have a traceback.

```suggestion
            try:
                raise ValueError("Something went wrong")
            except ValueError as e:
                span.record_exception(e)
```

src/appsignal/cli.py
```diff
@@ -135,3 +150,54 @@ def _write_file(self):
...
+            name=self._name,
+            push_api_key=self._push_api_key,
+            log_level="trace",
+        )
+        print(f"Starting AppSignal client for {self._name}...")
+        client.start()
+
+        tracer = trace.get_tracer(__name__)
+
+        # Performance sample
```
tombruijn, line 182: Please add the `demo_sample: true` tag to both samples so the UI will show a special box explaining it's a demo sample. See also [how we do it in Node.js](https://github.com/appsignal/appsignal-nodejs/blob/484e3a161211a2d7d789a52b1efa81df2719d8a9/src/demo.ts#L26).

src/appsignal/cli.py
tombruijn: > no need to ask for it via input.

In my opinion, none of these options need to support a prompt for this command. Keep it if it's easier, but otherwise remove it.

unflxw: Do note that we're currently not reading from the project's config file to get these values (this is the same issue that @luismiramirez just fixed for Node.js)

tombruijn (approved):

Issue #9, opened 2023-03-24
Implement Flask instrumentation support


unflxw:
A Flask [OpenTelemetry instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-flask) exists, and supporting it would be neat. If Django is Rails, then Flask is Sinatra: it's the "light-weight", DIY option for making a web application.

## To do

- [x] Create a test setup for the instrumentation https://github.com/appsignal/test-setups/pull/112
- [x] Write an extractor for the agent https://github.com/appsignal/appsignal-agent/pull/990
- [x] Add an instrumentation adder for the package https://github.com/appsignal/appsignal-python/pull/60
- [x] Document it for the package setup in PR https://github.com/appsignal/appsignal-docs/pull/1297/commits/6e90668a64690ba8a295d9bf28a45aacda95d4a6

unflxw:
@luismiramirez I notice now that the to do list says "document this for the existing stand-alone-based setup". We can still do that if we think it's valuable, but now that [the future Python package docs](https://github.com/appsignal/appsignal-docs/pull/1300) are ready, we might want to only write the docs for the Python package and add them to that PR.

(We also haven't really tested this with the stand-alone setup, I don't think -- which is also fine in my view)

tombruijn:
I suggest we write it for the new docs.

luismiramirez:
Yep, my idea is to add the instructions to [this PR](https://github.com/appsignal/appsignal-docs/pull/1297/files)

PR #60, opened 2023-05-23
Add Flask integration

luismiramirez:
Automatically start the Flask instrumentor included in the SDK when starting the AppSignal client.

Part of: https://github.com/appsignal/appsignal-python/issues/9

[skip changeset]

### DIFF
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -31,6 +31,12 @@ def add_django_instrumentation():
     DjangoInstrumentor().instrument(response_hook=response_hook)
 
 
+def add_flask_instrumentation():
+    from opentelemetry.instrumentation.flask import FlaskInstrumentor
+
+    FlaskInstrumentor().instrument()
+
+
 def add_jinja2_instrumentation():
     from opentelemetry.instrumentation.jinja2 import Jinja2Instrumentor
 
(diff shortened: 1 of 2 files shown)


unflxw:
I'm happy to make a separate issue for it, but I think it should be possible to add the params of the request relatively easily 👀 

The instrumentation supports request hooks: https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/flask/flask.html#request-response-hooks -- from that WSGI environ that the request hook takes, it should be possible to re-constitute a [Flask Request object](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Request) and [ask if it's contents are JSON](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Request.is_json) -- and if they are, [set the JSON contents](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Request.json) as the `appsignal.request.params` in the span.

src/appsignal/opentelemetry.py
```diff
@@ -31,6 +31,12 @@ def response_hook(span, request, response):
     DjangoInstrumentor().instrument(response_hook=response_hook)
 
 
+def add_flask_instrumentation():
+    from opentelemetry.instrumentation.flask import FlaskInstrumentor
+
+    FlaskInstrumentor().instrument(enable_commenter=True, commenter_options={})
```
tombruijn, line 37: Do we want to enable the sql commenter? We didn't for django. Could mess up the grouping of sanitized queries.

tombruijn:
> I'm happy to make a separate issue for it, but I think it should be possible to add the params of the request relatively easily 👀 (I have now jinxed it)

I'm okay with making a separate issue for it, to merge this PR quicker.

src/appsignal/opentelemetry.py
unflxw: Indeed, these sorts of comments used to mess up the grouping (Prisma adds them with no option to disable them) so [we changed `sql_lexer` to remove trailing comments](https://github.com/appsignal/sql_lexer/commit/d629f5a8f259e7cfeaf64aa4fb9142e62d59afc3).

So it shouldn't matter, but I agree it's still better to not send a comment just to remove it later.

tombruijn (approved):

unflxw (approved):

PR #65, opened 2023-05-30
Add query params to Flask spans

### LINKED ISSUE #62: Record Flask request parameters
The OpenTelemetry Flask integration supports [request hooks](https://github.com/appsignal/appsignal-docs/pull/1297/files). We should be able to build a [Flask Request] object from the WSGI environ present in the hooks and take the params if any to stamp them in the span.

## ToDo

- [x] Use the request hooks for Flask to stamp request parameters in the span

luismiramirez:
Parse the query params if present for each request and add them to the request parameters magic attribute.

Closes #62 

[skip changeset]

### DIFF
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -34,8 +34,17 @@ def add_django_instrumentation():
 
 def add_flask_instrumentation():
     from opentelemetry.instrumentation.flask import FlaskInstrumentor
+    import json
+    from urllib.parse import parse_qs
 
-    FlaskInstrumentor().instrument()
+    def request_hook(span, environ):
+        if span and span.is_recording():
+            query_params = parse_qs(environ.get("QUERY_STRING", ""))
+            span.set_attribute(
+                "appsignal.request.parameters", json.dumps({"args": query_params})
+            )
+
+    FlaskInstrumentor().instrument(request_hook=request_hook)
 
 
 def add_jinja2_instrumentation():


src/appsignal/opentelemetry.py
```diff
@@ -34,8 +34,17 @@ def response_hook(span, request, response):
...
 def add_flask_instrumentation():
     from opentelemetry.instrumentation.flask import FlaskInstrumentor
+    import json
+    from urllib.parse import parse_qs
+
+    def request_hook(span, environ):
+        if span and span.is_recording():
+            query_params = parse_qs(environ.get("QUERY_STRING", ""))
+            span.set_attribute(
+                "appsignal.request.parameters", json.dumps({"GET": query_params})
```
unflxw, line 44: Why do we write the query params as `{"GET": ...}`? What does the "GET" mean here?

tombruijn: I think it's a copy from what I did with the [Django instrumentation](https://github.com/appsignal/appsignal-python/blob/e84d4752bfe6a6857755f417084bae150c2d4a7f/src/appsignal/opentelemetry.py#L29). It adds a `GET` and `POST` key. Might be useful to differentiate between the types of params if we ever make the post payload thing work.

unflxw: Ah, I see! I do like the idea of differentiating between query parameters and body content (instead of merging them together, like we do in other integrations)

I'm not so sure about using the terms "GET" and "POST" -- a POST request can have query ("GET") parameters. But that's just nitpicking. 👍

luismiramirez: IIRC we use it as a way of differentiating query params from body params?

tombruijn: > IIRC we use it as a way of differentiating query params from body params?

Yes

> I'm not so sure about using the terms "GET" and "POST" -- a POST request can have query ("GET") parameters. But that's just nitpicking. 👍

Agreed. Better naming suggestions are welcome :)

unflxw: For Django, `GET` and `POST` map to the `request.GET` and `request.POST` methods that you obtain them from. I think it's good to use a naming that will be familiar to the framework's users, even if that naming is itself not great.

For Flask, `request.query_string` or `request.args` is how users would access the query string parameters, so I'd consider "query" or "args". And in the future, for the body parsed as JSON (which would be accessed via `request.json`) I'd consider "body" or "json".

luismiramirez: I'm going to use `args` and merge this PR

PR #66, opened 2023-06-01
Validate Push API key when installing

### LINKED ISSUE #42: Validate Push API key in install command
Split from https://github.com/appsignal/appsignal-python/pull/38#discussion_r1184674320.

To do
===

- [ ] Validate the Push API key inputted in the install command (or via command line)
- [ ] Respect HTTP proxy and CA file path configuration values

luismiramirez:
When the install command is run and the user inputs an API key, it makes a request to AppSignal to validate it. If it's not valid, the installer aborts with a message.

Closes #42

[skip changeset]

### DIFF
--- a/src/appsignal/cli.py
+++ b/src/appsignal/cli.py
@@ -1,11 +1,13 @@
 from abc import ABC, abstractmethod
 import os
 import json
+import requests
 
 from typing import Optional
 from docopt import docopt
 
 from .__about__ import __version__
+from .config import Config
 
 from appsignal.client import Client
 from opentelemetry import trace
@@ -87,6 +89,7 @@ class InstallCommand(AppsignalCLICommand):
     def __init__(self, push_api_key=None):
         self._push_api_key = push_api_key
         self._name = None
+        self._config = Config()
 
     def run(self) -> int:
         print("👋 Welcome to the AppSignal for Python installer!")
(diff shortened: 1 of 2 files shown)


luismiramirez:
I kept the HTTP proxy and the CA file out of this PR. Based on our other installers, that's something to take into account during the diagnose command, which is not present yet in this integration.

unflxw:
> I kept the HTTP proxy and the CA file out of this PR. Based on our other installers, that's something to take into account during the diagnose command

I'm happy to keep it out for now and merge this, but I'm not sure what you mean about it being for the diagnose command?

For example, the Ruby integration's install command calls the [Appsignal::AuthCheck](https://github.com/appsignal/appsignal-ruby/blob/main/lib/appsignal/cli/install.rb#L44) with [a config](https://github.com/appsignal/appsignal-ruby/blob/main/lib/appsignal/cli/install.rb#L307-L309) (which reads the environment variables) and that in turn uses the [Appsignal::Transmitter](https://github.com/appsignal/appsignal-ruby/blob/main/lib/appsignal/transmitter.rb#L11), which uses the [HTTP proxy](https://github.com/appsignal/appsignal-ruby/blob/main/lib/appsignal/transmitter.rb#L72) and [CA file path](https://github.com/appsignal/appsignal-ruby/blob/main/lib/appsignal/transmitter.rb#L83) configuration to make the request.

src/appsignal/cli.py
```diff
@@ -215,6 +225,11 @@ def _add_dependency(self, dependency_name):
             with open(requirement_file, "a") as f:
                 f.write(f"{dependency_name}\n")
 
+    def _validate_push_api_key(self):
+        url = f"https://push.appsignal.com/1/auth?api_key={self._push_api_key}"
```
tombruijn, line 229: Please use the config's `endpoint` key for the domain here.

tombruijn (approved):
Looks good so far, but you can pick up the endpoint, ca cert and proxy config next. No need to wait for the diagnose cli.

PR #67, opened 2023-06-01
Use shell to remove agent in tests

luismiramirez:
Using `shutil.rmtree` might be the cause of a race condition that makes the test suite fail on CI. Shelling out with a plain `rm -rf` addresses the problem.

[skip changeset]

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,7 +2,6 @@ import os
 import logging
 import pytest
 import platform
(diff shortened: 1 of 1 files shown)


unflxw:
I did quite a number of re-runs and I haven't seen the flaky error pop up again, so I think we can assume it's fixed. 🤞

PR #71, opened 2023-06-07
Fix certificate error in installer

unflxw:
Make sure that the `cacert.pem` file is added to the package.

While we're at it, remove the unit tests from the package, and make sure `py.typed` is added as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,8 +50,12 @@ path = "src/appsignal/__about__.py"
 
 [tool.hatch.build]
 sources = ["src"]
-exclude = ["src/scripts"]
-artifacts = ["src/appsignal/appsignal-agent"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -50,8 +50,12 @@ path = "src/appsignal/__about__.py"
 
 [tool.hatch.build]
 sources = ["src"]
-exclude = ["src/scripts"]
-artifacts = ["src/appsignal/appsignal-agent"]
+exclude = ["src/scripts", "tests"]
```
tombruijn, line 53:
```suggestion
exclude = ["src/scripts", "tests", "conftest.py"]
```
This test config file seems to be included as well.

tombruijn (changes requested):

tombruijn (approved):

PR #73, opened 2023-06-08
Allow custom agent builds

unflxw:
[skip changeset] because this is all build script related

### [Remove the agent cache on clean](https://github.com/appsignal/appsignal-python/commit/dc6f0809ce7bec61a79529e858d1c74e3aa74af9)

When `mono clean` is invoked, the agent binaries should be
downloaded again, in case the agent binary cache was tampered with.

### [Exit if tarball has no agent binary](https://github.com/appsignal/appsignal-python/commit/1b5ae8b2543eae1b1af70e524b712076f402229a)

Not sure how this would happen, but we should error if it does.

### [Allow custom agent builds](https://github.com/appsignal/appsignal-python/commit/d5557b94f545b0ced2e1539f8d900fd3371ba762)

An argument can be passed to `hatch run build:me` specifying the
path to the agent: `hatch run build:me path/to/agent`.

Additionally, you can use the `--keep-agent` flag to use the current
agent binary that is already present at `src/appsignal/appsignal-agent`
from a previous build: `hatch run build:me --keep-agent`.

The latter can be useful in containerised environments like our test
setups, which mount the integration repository as a volume in a
container, but not the agent repository -- meaning the location of the
agent artifact is not accessible when the integration package is
being built.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,14 @@ hatch run build:me # for your current platform
 hatch run build:for <triple> # for a specific agent triple
 ```
 
+#### Custom agent build
+```sh
+hatch run build:me /path/to/agent
+# or place the desired agent binary at
+# `src/appsignal/appsignal-agent`, and then:
+hatch run build:me --keep-agent
+```
+
 ### Clean up build artifacts
 ```sh
 hatch clean # clean dist folder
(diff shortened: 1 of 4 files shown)


unflxw:
To do: add it to the README. (oops!)

PR #82, opened 2023-06-28
Annotate None return type

orsinium:
I noticed that quite a few type annotations are missing. This PR adds the easiest ones: None return type. I'll bring more after this PR is merged.

### DIFF
--- a/src/appsignal/agent.py
+++ b/src/appsignal/agent.py
@@ -14,17 +14,17 @@ def is_agent_active():
     return agent_active
 
 
-def _reset_agent_active():
+def _reset_agent_active() -> None:
     global agent_active
     agent_active = False
 
 
-def agent_is_active():
+def agent_is_active() -> None:
     global agent_active
     agent_active = True
 
 
-def start_agent(config: Config):
+def start_agent(config: Config) -> None:
     config.set_private_environ()
     p = subprocess.Popen(
         [AGENT_PATH, "start", "--private"],
--- a/src/appsignal/cli/command.py
+++ b/src/appsignal/cli/command.py
@@ -11,13 +11,13 @@ class AppsignalCLICommand(ABC):
     _push_api_key: Optional[str]
     _name: Optional[str]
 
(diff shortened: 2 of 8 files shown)


unflxw (approved):
Thank you for your contribution!

If I recall correctly, the type checker won't check functions unless their signature has type annotations, right? Do you know if there's a way to change that, so that they're checked regardless?

tombruijn (approved):

orsinium:
> If I recall correctly, the type checker won't check functions unless their signature has type annotations, right? 

By default, yes.

> Do you know if there's a way to change that, so that they're checked regardless?

It's controlled by `check_untyped_defs` option for mypy, you already have it.

There are more reasons to annotate the None return type explicitly rather than just encourage mypy to type check the function body. Let's take these 2 functions as an example:

```python
def create_user():
  pass
```

```python
def create_user() -> None:
  pass
```

1. The default return type is `Any`, not `None`. So, if somewhere in the code we have `created = create_user()`, the first definition will pass just fine, but for the second definition, mypy will tell you that you try to use the result of a function that returns nothing.
1. As a reader, I'll know that you haven't just forgotten to annotate the function. For example, in reviews, I always nudge people to annotate the functions they introduced. Eventually, I want to adjust the mypy config to require type annotations for all functions.

This behavior is explicitly defined in the typing PEP:

> It is recommended but not required that checked functions have annotations for all arguments and the return type. For a checked function, the default annotation for arguments and for the return type is Any.
> https://peps.python.org/pep-0484/#the-meaning-of-annotations

unflxw:
Thank you for the explanation @orsinium! Makes perfect sense. Coming back to Python after quite some time, I'm really happy to see the community embracing typing, but I'm pretty lost on the details of the tooling.

> Eventually, I want to adjust the mypy config to require type annotations for all functions.

I think that would be the ideal scenario, yes! I find type checking much more useful when I know that it _will_ complain if I do something wrong -- not just that it _might_.

PR #83, opened 2023-06-28
Mypy config

orsinium:
1. Use config for mypy options, so that the IDE and `hatch run lint:typing` produce the same results. The `--enable-incomplete-feature` flag statys, though, I don't think there is a config option for that.
1. Resolve mypy violation by importing Unpack from typing_extension. You don't need to install typing_extension for that to work, myp ships with stubs for it.
1. Explicitly list mypy dependencies instead of using `--install-types`. Let your package manager deal with packages. It's faster, it's explicit, it's reliable.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,10 +82,13 @@ dependencies = [
   "opentelemetry-sdk",
   "opentelemetry-exporter-otlp-proto-http",
   "docopt",
+  "types-deprecated",
+  "types-docopt",
+  "types-requests",
 ]
 
 [tool.hatch.envs.lint.scripts]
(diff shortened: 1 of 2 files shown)


unflxw (approved):
Thanks for the contribution! 🎉

I take it that `--non-interactive` was redundant, once you remove `--install-types`.

tombruijn (approved):

orsinium:
> I take it that `--non-interactive` was redundant, once you remove `--install-types`.

That's correct. The only interactive thing in mypy is when it asks for a confirmation before installing missed packages if `--install-types` is specified, that's it. Without the flag, if you add a dependency that requires stubs to be installed, mypy will report it as a regular type error, without any interactive prompts.

PR #85, opened 2023-06-29
Sort all imports using isort

orsinium:
[isort](https://pycqa.github.io/isort/) is a code formatter for sorting and grouping python imports. The PR add isort into `lint` dependencies and uses it to sort all imports.


unflxw (approved):
Thanks once again! 💯

PR #86, opened 2023-06-29
Improve coverage report

orsinium:
1. Exclude `tests` and `src/scripts` from the coverage report. Calculate coverage only for `src/appsignal`.
1. Set minimal coverage.
1. Calculate coverage even when running pytest without any options.
1. Produce HTML coverage report.
1. Remove undocumented `cov` and `no-cov` hatch commands in favor of `test:pytest` used on CI and described in the README.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __pycache__
 /dist
 /src/appsignal/appsignal-agent
 /tmp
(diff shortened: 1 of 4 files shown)


unflxw (approved):
Thank you!

PR #89, opened 2023-06-30
Diagnose command

jeffkreeftmeijer:
This PR includes a diagnose command which can be called through `appsignal diagnose`, a report that can be submitted to AppSignal, and a report that's printed to the terminal. 

This also involves some shortcuts being made so we can work from a passing test suite. So far, the shortcuts made are a couple of hardcoded values, which will be fixed in this pull request, but are moved back for now. I'm listing them here so we're clear on what's not yet implemented, outside of the test failures:

# TODO

- [x] Add the `diagnose` command
- [x] Make diagnose tests pass -> https://github.com/appsignal/diagnose_tests/pull/92
- [x] Add diagnose tests to CI

### DIFF
--- a/src/appsignal/cli/diagnose.py
+++ b/src/appsignal/cli/diagnose.py
@@ -1,9 +1,12 @@
+# ruff: noqa: E501
+
 import json
 import os
 import platform
 import urllib
 from argparse import ArgumentParser
 from pathlib import Path
+from typing import Any, Optional
 
 import requests
 
@@ -16,59 +19,55 @@ from .command import AppsignalCLICommand
 
 
 class AgentReport:
-    def __init__(self, report):
+    def __init__(self, report: dict) -> None:
         self.report = report
 
-    def extension_loaded(self):
+    def extension_loaded(self) -> str:
         return self.report["boot"]["started"]["result"]
 
-    def configuration_valid(self):
+    def configuration_valid(self) -> str:
         if self.report["config"]["valid"]["result"]:
             return "valid"
-        else:
-            return "invalid"
+        return "invalid"
 
-    def started(self):
+    def started(self) -> str:
         if self.report["boot"]["started"]["result"]:
             return "started"
-        else:
-            return "not started"
+        return "not started"
 
-    def user_id(self):
+    def user_id(self) -> str:
         return self.report["host"]["uid"]["result"]
 
-    def group_id(self):
+    def group_id(self) -> str:
         return self.report["host"]["gid"]["result"]
 
-    def logger_started(self):
+    def logger_started(self) -> str:
         if "logger" in self.report and self.report["logger"]["started"]["result"]:
             return "started"
-        else:
-            return "not started"
+        return "not started"
 
-    def working_directory_user_id(self):
+    def working_directory_user_id(self) -> bool:
         return (
             "working_directory_stat" in self.report
             and self.report["working_directory_stat"]["uid"]["result"]
         )
 
-    def working_directory_group_id(self):
+    def working_directory_group_id(self) -> bool:
         return (
             "working_directory_stat" in self.report
             and self.report["working_directory_stat"]["gid"]["result"]
         )
 
-    def working_directory_permissions(self):
+    def working_directory_permissions(self) -> bool:
         return (
             "working_directory_stat" in self.report
             and self.report["working_directory_stat"]["mode"]["result"]
         )
 
-    def lock_path(self):
+    def lock_path(self) -> str:
         if "lock_path" in self.report and self.report["lock_path"]["created"]["result"]:
             return "writable"
-        else:
-            return "not writable"
+        return "not writable"
 
 
 class DiagnoseCommand(AppsignalCLICommand):
@@ -85,7 +84,7 @@ class DiagnoseCommand(AppsignalCLICommand):
             help="Do not send the report to AppSignal",
         )
 
-    def run(self):
+    def run(self) -> int:
         agent = Agent()
         self.config = Config()
         self.send_report = self.args.send_report
@@ -144,7 +143,9 @@ class DiagnoseCommand(AppsignalCLICommand):
         if self.no_send_report:
             print("Not sending report. (Specified with the --no-send-report option.)")
 
-    def _header(self):
+        return 0
+
+    def _header(self) -> None:
         print("AppSignal diagnose")
         print("=" * 80)
         print("Use this information to debug your configuration.")
@@ -153,16 +154,16 @@ class DiagnoseCommand(AppsignalCLICommand):
         print("Send this output to support@appsignal.com if you need help.")
         print("=" * 80)
 
-    def _library_information(self):
-        library_report = self.report["library"]
+    def _library_information(self) -> None:
+        library_report: Any = self.report["library"]
         print("AppSignal library")
         print("  Language: Python")
         print(f'  Package version: "{library_report["package_version"]}"')
         print(f'  Agent version: "{library_report["agent_version"]}"')
         print(f'  Extension loaded: {library_report["extension_loaded"]}')
 
-    def _host_information(self):
-        host_report = self.report["host"]
+    def _host_information(self) -> None:
+        host_report: Any = self.report["host"]
         print("Host information")
         print(f'  Architecture: "{host_report["architecture"]}"')
         print(f'  Operating System: "{host_report["os"]}"')
@@ -170,7 +171,7 @@ class DiagnoseCommand(AppsignalCLICommand):
         print(f'  Root user: {host_report["root"]}')
         print(f'  Running in container: {host_report["running_in_container"]}')
 
-    def _agent_information(self):
+    def _agent_information(self) -> None:
         print("Agent diagnostics")
         print("  Extension tests")
         print(f"    Configuration: {self.agent_report.configuration_valid()}")
@@ -191,28 +192,28 @@ class DiagnoseCommand(AppsignalCLICommand):
         )
         print(f"    Lock path: {self.agent_report.lock_path()}")
 
-    def _configuration_information(self):
+    def _configuration_information(self) -> None:
         print("Configuration")
 
         for key in self.config.options:
-            print(f"  {key}: {repr(self.config.options[key])}")
+            print(f"  {key}: {self.config.options[key]!r}")  # type: ignore
 
         print()
         print("Read more about how the diagnose config output is rendered")
         print("https://docs.appsignal.com/python/command-line/diagnose.html")
 
-    def _validation_information(self):
-        validation_report = self.report["validation"]
+    def _validation_information(self) -> None:
+        validation_report: Any = self.report["validation"]
         print("Validation")
         print(f'  Validating Push API key: {validation_report["push_api_key"]}')
 
-    def _paths_information(self):
+    def _paths_information(self) -> None:
         print("Paths")
 
-    def _report_information(self):
+    def _report_information(self) -> None:
         print("Diagnostics report")
 
-    def _report_prompt(self):
+    def _report_prompt(self) -> bool | None:
         print("  Do you want to send this diagnostics report to AppSignal?")
         print("  If you share this report you will be given a link to")
         print("  AppSignal.com to validate the report.")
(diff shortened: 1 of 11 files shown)


src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+                "os": platform.system().lower(),
+                "os_distribution": self._os_distribution(),
+                "root": False,
+                "running_in_container": self.config.option("running_in_container")
+                is not None,
+            },
+            "library": {
+                "language": "python",
+                "package_version": __version__,
+                "agent_version": "91f1a7c",
```
unflxw, line 131: This value is still hardcoded. [There was a conversation about adding this version number to the agent's diagnose output](https://appsignal.slack.com/archives/CNPP953E2/p1688044462438079), as the agent is already aware of its own version number.

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+
+    def _report_prompt(self) -> bool | None:
+        print("  Do you want to send this diagnostics report to AppSignal?")
+        print("  If you share this report you will be given a link to")
+        print("  AppSignal.com to validate the report.")
+        print("  You can also contact us at support@appsignal.com")
+        print("  with your support token.")
+        print()
+
+        send_report = input("  Send diagnostics report to AppSignal? (Y/n):   ")
```
unflxw, line 270:
```suggestion
        send_report = input("  Send diagnostics report to AppSignal? (Y/n):   ").lower()
```

SInce we suggest capital `Y` for yes to indicate that it's the default, I think its best if we accept the inputs in a case-independent way.

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+            action="store_true",
+            help="Do not send the report to AppSignal",
+        )
+
+    def run(self) -> int:
+        agent = Agent()
+        agent_json = json.loads(agent.diagnose())
+        self.config = Config()
+        self.send_report = self.args.send_report
+        self.no_send_report = self.args.no_send_report
```
unflxw, line 98: I think it's worth throwing an error and bailing out early if both `--send-report` and `--no-send-report` are specified. Otherwise, we'll do weird things later (e.g. printing both "Sending this report" and "Not sending this report" messages)

Additionally, note that `argparse` has this `BooleanOptionalAction` that implements `--option` and `--no-option` for you automatically. Unfortunately, if I'm reading the docs right, it's only implemented in Python 3.9 and above. :(

(You'll have to search for it manually [in the `add_argument()` docs](https://docs.python.org/3/library/argparse.html#action), I can't link to it directly because Python's documentation is entirely composed of unstructured streams of consciousness)

unflxw (commented):
Holding the approve on the hardcoded agent version value. Looks good otherwise.

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+            if os.path.isdir(path):
+                return "directory"
+        return "Path does not exist"
+
+    def _file_is_writable(self, path: str) -> bool:
+        return os.access(path, os.W_OK)
+
+    def _os_distribution(self) -> str:
+        try:
+            return platform.freedesktop_os_release()["NAME"]
```
tombruijn, line 379: This gives me a different error than is potentially expected here:

```
$ hatch run appsignal diagnose
Traceback (most recent call last):
  File "/Users/tombruijn/Library/Application Support/hatch/env/virtual/appsignal-beta/BAIuE2ZA/appsignal-beta/bin/appsignal", line 10, in <module>
    sys.exit(run())
  File "/Users/tombruijn/appsignal/python/src/appsignal/cli/base.py", line 24, in run
    sys.exit(main(sys.argv[1:]))
  File "/Users/tombruijn/appsignal/python/src/appsignal/cli/base.py", line 38, in main
    return cmd.run()
  File "/Users/tombruijn/appsignal/python/src/appsignal/cli/diagnose.py", line 123, in run
    "os_distribution": self._os_distribution(),
  File "/Users/tombruijn/appsignal/python/src/appsignal/cli/diagnose.py", line 379, in _os_distribution
    return platform.freedesktop_os_release()["NAME"]
AttributeError: module 'platform' has no attribute 'freedesktop_os_release'
```

src/appsignal/cli/diagnose.py
tombruijn, line 270:
```suggestion
        send_report = input("  Send diagnostics report to AppSignal? (Y/n): ")
```

Extra spaces unnecessary?

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+                "sources": self.config.sources,
+            },
+            "host": {
+                "architecture": platform.machine(),
+                "heroku": os.environ.get("DYNO") is not None,
+                "language_version": platform.python_version(),
+                "os": platform.system().lower(),
+                "os_distribution": self._os_distribution(),
+                "root": False,
+                "running_in_container": self.config.option("running_in_container")
```
tombruijn, line 125: Since the current value is based on what the user has configured, rather than what we determine by reading the host environment, I suggest we omit it from this section. [This section is only meant for what we detect](https://github.com/appsignal/integration-guide/blob/main/features/cli/diagnose.md#host-report), which allows us to compare what we detected against what the user configured. How it's set up now could create confusion by us thinking this is what the package detected.

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+                "options": self.config.options,
+                "sources": self.config.sources,
+            },
+            "host": {
+                "architecture": platform.machine(),
+                "heroku": os.environ.get("DYNO") is not None,
+                "language_version": platform.python_version(),
+                "os": platform.system().lower(),
+                "os_distribution": self._os_distribution(),
+                "root": False,
```
tombruijn, line 124:
```suggestion
                "root": os.getuid() == 0,
```

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,390 @@
...
+
+    def _paths_data(self) -> dict:
+        working_dir = os.getcwd() or ""
+        log_file_path = self.config.log_file_path() or ""
+        log_path = self.config.option("log_path") or os.path.dirname(log_file_path)
+
+        try:
+            with open(log_file_path) as log_file:
+                lines = log_file.readlines()
+                log_file_contents = lines[-10:]
```
tombruijn, line 317: This only stores the last 10 lines of the file in the diagnose report JSON. We would like to [read up to 2MB of data](https://github.com/appsignal/appsignal-ruby/blob/5ddde58bb492984626d2dbddb292cecdfc225576/lib/appsignal/cli/diagnose/paths.rb#L74-L77).

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,388 @@
...
+        except FileNotFoundError:
+            return {}
+
+    def _file_type(self, path: str) -> str:
+        if os.path.exists(path):
+            if os.path.isfile(path):
+                return "file"
+            if os.path.isdir(path):
+                return "directory"
+        return "Path does not exist"
```
tombruijn, line 370: This value will also be used for the JSON report. It would be better to leave this field empty when the path does not exist and only show this label in the CLI output.

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,388 @@
...
+                "mode": self._file_stat(working_dir).get("mode"),
+                "ownership": self._file_stat(working_dir).get("ownership"),
+                "path": working_dir,
+                "type": self._file_type(working_dir),
+                "writable": self._file_is_writable(working_dir),
+            },
+        }
+
+    def _file_exists(self, path: str) -> bool:
+        return os.path.isfile(path) or os.path.isdir(path)
```
tombruijn, line 349: Would `os.path.exists` also work here? Like is used in the `_file_type` method.

luismiramirez: Yup it should. This is the result of me starting to use the `Path` module and then switching its usage to `os` 🙃

src/appsignal/cli/diagnose.py
luismiramirez: That's what the diagnose-tests expect:

```
"  Send diagnostics report to AppSignal? (Y/n):   Transmitting diagnostics report",
```

src/appsignal/cli/diagnose.py
```diff
@@ -0,0 +1,382 @@
...
+                is not None,
+            },
+            "library": {
+                "language": "python",
+                "package_version": __version__,
+                "agent_version": "91f1a7c",
+            },
+            "paths": self._paths_data(),
+            "process": {
+                "pid": os.getpid(),
```
tombruijn, line 132:
```suggestion
                "uid": os.getuid(),
```
We want to track the user id here.

src/appsignal/cli/diagnose.py
luismiramirez: It looks like it was added from 3.10 but it's a strange thing because some older versions of Python 3 seem to have that method available. Could it be that your `hatch run appsignal diagnose` is using a < 3.10 python?

Anyway, being that the case, we need to add another rescue for `AttributeError`

src/appsignal/cli/diagnose.py
```diff
@@ -309,18 +303,11 @@ def _paths_data(self) -> dict:
...
         log_path = self.config.option("log_path") or os.path.dirname(log_file_path)
-
-        try:
-            with open(log_file_path) as log_file:
-                lines = log_file.readlines()
-                log_file_contents = lines[-10:]
-                log_file_contents = [line.strip() for line in log_file_contents]
-        except FileNotFoundError:
-            log_file_contents = [""]
+        log_file_lines = self._last_ten_lines(log_file_path)
```
tombruijn, line 303:
```suggestion
        log_file_lines = self._read_last_two_mib(log_file_path)
```

We want to include the last 2MB of the log file in the JSON report, but only show the last 10 lines in the CLI output.

tombruijn:
I made some fixes in this PR https://github.com/appsignal/appsignal-python/pull/108
along with this diagnose_tests PR: https://github.com/appsignal/diagnose_tests/pull/93

tombruijn (approved):

unflxw (approved):
I thouight I'd approved this before; my bad!

src/appsignal/cli/diagnose.py
```diff
@@ -373,7 +373,10 @@ def _file_is_writable(self, path: str) -> bool:
     def _os_distribution(self) -> str:
         try:
             return platform.freedesktop_os_release()["NAME"]
-        except OSError:
+        # The platform.freedesktop_os_release is not available in all Python
+        # versions that we support. That's why we catch AttributeError along with
+        # OSError which is thrown when the host OS doesn't comply with freedesktop.
+        except (OSError, AttributeError):
```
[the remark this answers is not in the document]
tombruijn: It reads the same file, but it would be nice to include the whole function return value and make the formatting the same.

```
$ docker run --rm -it python
>>> import platform
>>> platform.freedesktop_os_release()
{'NAME': 'Debian GNU/Linux', 'ID': 'debian', 'PRETTY_NAME': 'Debian GNU/Linux 11 (bullseye)', 'VERSION_ID': '11', 'VERSION': '11 (bullseye)', 'VERSION_CODENAME': 'bullseye', 'HOME_URL': 'https://www.debian.org/', 'SUPPORT_URL': 'https://www.debian.org/support', 'BUG_REPORT_URL': 'https://bugs.debian.org/'}

$ docker run --rm -it python bash
root@809d031cbba5:/# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
The second example being how we send it for other integrations.

PR #90, opened 2023-07-01
Upgrade type annotations

orsinium:
1. Add `from __future__ import annotations` in every Python module. Doing so will make Python not evaluate type annotations at runtime.
2. Upgrade all type annotations. Since all type annotations are now lazy, we can use the latest syntax for them. So, instead of `Optional[List[Union[str, int]]]` we can write `list[str | int] | None`.

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -3,16 +3,7 @@ from __future__ import annotations
 import os
 import platform
 import tempfile
-from typing import (
-    TYPE_CHECKING,
-    List,
-    Literal,
-    Optional,
-    TypedDict,
-    Union,
-    cast,
-    get_args,
-)
+from typing import TYPE_CHECKING, List, Literal, TypedDict, cast, get_args
 
 from .__about__ import __version__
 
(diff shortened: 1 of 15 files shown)


tombruijn (approved):

src/appsignal/config.py
```diff
@@ -3,16 +3,7 @@
...
-    TYPE_CHECKING,
-    List,
-    Literal,
-    Optional,
-    TypedDict,
-    Union,
-    cast,
-    get_args,
-)
+from typing import TYPE_CHECKING, List, Literal, TypedDict, cast, get_args
```
unflxw, line 6: Asking in order to better understand how type checking works: I see that, while most annotations have been changed to use `list[...]`, we still import `List` here, so that we can use it in expressions like `cast(List[DefaultInstrumentation]`.

Is it that we can't use `cast(list[DefaultInstrumentation])` because the call to `cast` happens in the run-time code, not in annotations (even if it's a no-op at run-time) and `list[...]` is not valid syntax at run-time in (older?) versions of Python?

And, because you've added `from __future__ import annotations`, the Python interpreter no longer checks what's going on within annotations, which allows us to use the newer `list[...]` form there. Did I get that right?

unflxw (approved):
Thank you! 🙌

src/appsignal/config.py
orsinium: Precisely! That's exactly what happens. Other examples of type annotations evaluated at runtime include type aliases, `bind` argument for `TypeVar`, base classes like `Generic` and their arguments, and pydantic models.

PR #91, opened 2023-07-01
Refactoring: turn agent wrapper into a singleton

orsinium:
Before: agent state is a global variable
managed from a set of functions
using `global`.

After: the agent state is managed by
Agent singleton. The state is incapsulated,
and all related logic shares the same namespace.
The resulting code is shorter and easier to test.

I'd prefer to avoid a global state altogether
and let Client manage the Agent,
but that would require to rethink the tests
that currently implicitly manage
the agent state from autouse fixtures.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,7 +5,7 @@ import tempfile
 
 import pytest
 
-from appsignal.agent import _reset_agent_active
+from appsignal.agent import agent
 
 
 @pytest.fixture(scope="function", autouse=True)
@@ -29,7 +29,7 @@ def remove_logging_handlers_after_tests():
 
 @pytest.fixture(scope="function", autouse=True)
 def reset_agent_active_state():
-    _reset_agent_active()
+    agent.active = False
 
 
 @pytest.fixture(scope="function", autouse=True)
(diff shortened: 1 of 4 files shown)


orsinium:
I started to wrap commit message bodies, and GitHub gives it away 🙃

tombruijn (approved):

tombruijn:
> I started to wrap commit message bodies, and GitHub gives it away 🙃

I think GitHub detects it's wrapped but not to the "standard" 72 characters so it assumes it's wrapped for a reason like this. I've seen it unwrap lines wrapped to 72 characters on my PRs automatically.

For example, when I would submit your PR with the format like this:

> Before: agent state is a global variable managed from a set of functions
> using global.
> 
> After: the agent state is managed by Agent singleton. The state is
> incapsulated, and all related logic shares the same namespace.  The
> resulting code is shorter and easier to test.
> 
> I'd prefer to avoid a global state altogether and let Client manage the
> Agent, but that would require to rethink the tests that currently
> implicitly manage the agent state from autouse fixtures.

unflxw:
Thanks once again. Merging!

> I'd prefer to avoid a global state altogether and let Client manage the Agent

We could implement it like that, but in practice, both the client and the agent must be _de facto_ singletons, as the agent process itself will kill any other agent processes with a similar enough configuration.

PR #94, opened 2023-07-03
Update OpenTelemetry HTTP exporter path

tombruijn:
Specify the OpenTelemetry HTTP exporter should export to `/v1/traces` path so we can more easily differentiate different requests on the HTTP endpoint on the agent.

No agent changes are needed with this agent version, it listens to any and all paths (it ignores the path).

[skip changeset]

### DIFF
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -98,7 +98,7 @@ def start_opentelemetry(config: Config) -> None:
 
     provider = TracerProvider()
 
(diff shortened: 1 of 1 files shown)


unflxw (approved):
I'm happy either way, but not sure if this needs a changelog entry, as the OpenTelemetry exporter path is not configurable and goes between two internal components of the integration. 🤷

tombruijn:
Fair point, removing it.

PR #95, opened 2023-07-04
Annotate everything

orsinium:
1. Add all missed type annotations.
1. Require all functions to be annotated. Now, if an annotation is missed, mypy will fail.

### DIFF
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 import os
-from typing import Callable, Dict
+from typing import TYPE_CHECKING, Callable, Mapping
 
 from opentelemetry import trace
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
@@ -10,6 +10,10 @@ from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 
 from .config import Config, list_to_env_str
+
+
+if TYPE_CHECKING:
+    from opentelemetry.trace.span import Span
 
 
 def add_celery_instrumentation() -> None:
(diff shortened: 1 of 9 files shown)


src/appsignal/opentelemetry.py
```diff
@@ -107,15 +110,16 @@ def start_opentelemetry(config: Config) -> None:
 
 
 def add_instrumentations(
-    config: Config, default_instrumentation_adders=DEFAULT_INSTRUMENTATION_ADDERS
+    config: Config,
+    adders: Mapping[str, DefaultInstrumentationAdder] = DEFAULT_INSTRUMENTATION_ADDERS,
```
orsinium, line 114: I simplified annotation because we want to let users specify their own opentelemetry instrumentation, even not explicitly supported by appsignal. At least that's how I interpret the purpose of this argument.

I also renamed it. I removed `default_` because you can pass not only the default adders but custom ones, and removed `instrumentation_` because the function name already says it.

unflxw: I don't think users would call this directly, with custom adders or otherwise. This is for our internal use -- the default instrumentation adders are only passed as a parameter [in order to make it easier to test this function's behavior](https://github.com/appsignal/appsignal-python/blob/1b07fe2b8add104d69f7d6f43f12a2016a702e6f/tests/test_opentelemetry.py#L22-L53). Perhaps the argument should have had an `_` prefix to make that clearer.

If users want to add their own OpenTelemetry instrumentation, they can do so already, without interacting with this function -- the ["instrumentation adders"](https://github.com/appsignal/appsignal-python/blob/1b07fe2b8add104d69f7d6f43f12a2016a702e6f/src/appsignal/opentelemetry.py#L15-L18) don't have an explicit dependency on AppSignal or its OpenTelemetry setup. This is to say, users can already do the following, at the bottom of `__appsignal__.py` or at the top of their application:

```python
from some.arbitrary.package import WhateverInstrumentor
WhateverInstrumentor.instrument()
```

The purpose of these instrumentation adders being named is so that they can be disabled using the `disable_default_instrumentations` config option -- and because the instrumentation's names are a bit tricky, I thought they benefitted from strong typing using literal, so users would get a warning if they typo'ed the config.

unflxw (approved):
Thank you! 🙌

tombruijn (approved):

PR #96, opened 2023-07-05
Restore literal typing for instrumentations

unflxw:
A nitpick on top of #95, restoring typing for the default instrumentation adders. Trying to prevent typos when referring to the default instrumentations (including our own typos!)

@orsinium I'm curious as to whether this makes sense to you, or am I too deep in the TypeScript mindset?

[skip changeset]

### DIFF
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -80,7 +80,9 @@ def add_requests_instrumentation() -> None:
 
 DefaultInstrumentationAdder = Callable[[], None]
 
-DEFAULT_INSTRUMENTATION_ADDERS = {
+DEFAULT_INSTRUMENTATION_ADDERS: Mapping[
+    Config.DefaultInstrumentation, DefaultInstrumentationAdder
+] = {
     "opentelemetry.instrumentation.celery": add_celery_instrumentation,
     "opentelemetry.instrumentation.django": add_django_instrumentation,
     "opentelemetry.instrumentation.flask": add_flask_instrumentation,
@@ -111,7 +113,9 @@ def start_opentelemetry(config: Config) -> None:
 
 def add_instrumentations(
     config: Config,
-    _adders: Mapping[str, DefaultInstrumentationAdder] = DEFAULT_INSTRUMENTATION_ADDERS,
+    _adders: Mapping[
+        Config.DefaultInstrumentation, DefaultInstrumentationAdder
+    ] = DEFAULT_INSTRUMENTATION_ADDERS,
 ) -> None:
     logger = logging.getLogger("appsignal")
     disable_list = config.options.get("disable_default_instrumentations") or []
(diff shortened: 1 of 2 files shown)


src/appsignal/opentelemetry.py
```diff
@@ -80,7 +80,9 @@ def add_requests_instrumentation() -> None:
 
 DefaultInstrumentationAdder = Callable[[], None]
 
-DEFAULT_INSTRUMENTATION_ADDERS = {
+DEFAULT_INSTRUMENTATION_ADDERS: Dict[
```
orsinium, line 83:
```suggestion
DEFAULT_INSTRUMENTATION_ADDERS: dict[
```

orsinium: Or better yet, use `Mapping`, then mypy will make sure nobody modifies the constant.

unflxw: 💡 I like that a lot. Will do, thanks!

orsinium:
Yeah, I don't mind the change. I relaxed the annotations when I thought we want to let users provide their own instrumentations.

I question, though, how some parts of the project are done. Usually, you can achieve better safety with less code with `Enum` and `dataclass` rather than `Literal` and `TypedDict`. I have a note for myself to look into it later.

orsinium:
Also, when I write similar code, I often use the "registry" pattern. Then it's easier to make sure that all adders are registered and is easier to test:

```python
class Registry:
  def instrument(self):
    ...

  def start(self):
    ...

  def add(self, adder):
    ...

registry = Registry()

@registry.add("opentelemetry.instrumentation.celery")
def add_celery_instrumentation() -> None:
  ...
```

unflxw:
> @orsinium Usually, you can achieve better safety with less code with Enum and dataclass rather than Literal and TypedDict

Please, feel free to look into it! The decisions that were made mostly come from my limited understanding of Python typing at the time that the project was started. (Admittedly, it hasn't gotten much less limited since then 😛)

That said, there are two things that we can do relatively easily with TypedDict, and _I think_ we wouldn't get them as easily from dataclasses:
- [Merging arbitrary config option dicts from several sources](https://github.com/appsignal/appsignal-python/blob/main/src/appsignal/config.py#L101-L113)
- [Accepting config options directly from `**kwargs` in the initialiser](https://github.com/appsignal/appsignal-python/blob/main/src/appsignal/client.py#L23-L24)

orsinium:
> That said, there are two things that we can do relatively easily with TypedDict

I know, I've seen how you use it :)

My idea for the Config would be to let users construct it and put the rest as defaults. Something like this:

```python
def get_env(env_var, default):
  def f():
    return os.environ.get(env_var, default)
  return f

@dataclass(frozen=True)
class Config:
  api_key = field(default_factory=get_env("APPSIGNAL_PUSH_API_ENDPOINT", "https://push.appsignal.com"))
```

And then it will be passed into Client explicitly:

```
config = Config(...)
client = Client(config)
```

That would make for users more apparent which env vars and which defaults are used for each option, give a good autocomplete, let config to be constructed in steps, and won't rely on still experimental and disabled by default support for Unpack.

tombruijn (approved):

PR #97, opened 2023-07-05
Migrate from docopt to argparse

### LINKED ISSUE #84: Get rid of docopt
The docopt library is needed only for the dev environment but it is an unconditional dependency which the appsignal SDK brings on the production. In Python, each dependency matters. 

An easy solution would be to introduce a `cli` extra for that. However, currently, docopt does nothing for the project that cannot be done by the good old argparse. So, my suggestion is to rewrite that little CLI that the project has to argparse and remove docopt from the dependencies. I'd be happy to take care of it, I'm sure I can migrate it without sacrificing anything. But since it's quite an invasive change (I might miss something and you might have reasons for using docopt), I need a green light from you first.

orsinium:
1. Remove docopt from dependencies.
1. Use argparse to configure CLI.
1. Use cached_property to ensure on type level that the API key and app name are always provided when we need them.

Before:

image

After:

image

TBH, I don't quite like that there are interactive prompts in the app. I'd remove them and make the CLI flags required. Then argparse will ask the user to provide these flags for us:

```
usage: appsignal demo [-h] --push-api-key PUSH_API_KEY --application APPLICATION
appsignal demo: error: the following arguments are required: --push-api-key, --application
```

Closes #84

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -124,7 +124,9 @@ def test_install_command_when_file_exists_no_overwrite(mocker):
     ):
         main(["install"])
 
-    # assert [] == cast(MagicMock, builtins.open).mock_calls
+    from appsignal.cli import install
+
+    assert install.open.mock_calls == []  # type: ignore[attr-defined]
 
 
 def test_install_comand_when_api_key_is_not_valid(mocker):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-import builtins
 from contextlib import contextmanager
-from typing import cast
 from unittest.mock import MagicMock
+
+from appsignal.cli.base import main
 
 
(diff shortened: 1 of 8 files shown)


tests/test_cli.py
```diff
@@ -131,32 +114,24 @@ def test_install_command_when_file_exists_no_overwrite(mocker):
...
+            "n",
+        ),
     ):
-        from appsignal.cli.install import InstallCommand
+        main(["install"])
 
-        InstallCommand().run()
-
-    assert [] == cast(MagicMock, builtins.open).mock_calls
+    # assert [] == cast(MagicMock, builtins.open).mock_calls
```
tombruijn, line 127: What should we do with this? Remove it?

tombruijn (approved):

tests/test_cli.py
orsinium: Good catch! Fixed.

orsinium:
For crying out loud!

PR #99, opened 2023-07-05
Make ruff configuration more strict

orsinium:
1. Enable all ruff rules that don't cause any trouble in the current code.
2. Resolve the reported violations.

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -3,7 +3,16 @@ from __future__ import annotations
 import os
 import platform
 import tempfile
-from typing import TYPE_CHECKING, Any, List, Literal, TypedDict, cast, get_args
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    ClassVar,
+    List,
+    Literal,
+    TypedDict,
+    cast,
+    get_args,
+)
 
 from .__about__ import __version__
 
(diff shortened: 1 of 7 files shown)


orsinium:
Damn, lintje failed.

tombruijn:
> Damn, lintje failed. Something about changeset?

The changeset one is a hint, but for this change it's not necessary. We use it to build our changelogs.

The error is that you made a commit without starting the subject with an uppercase letter. The rests are just hints.

src/appsignal/config.py
```diff
@@ -116,11 +125,11 @@ def option(self, option: str) -> Any:
         return self.options.get(option)
 
     @classmethod
-    def load_from_system(_cls) -> Options:
```
tombruijn, line 119: We can't prefix it with an underscore, even if we don't use it?

tombruijn (approved):

src/appsignal/config.py
orsinium: Yes. The convention is to always use `cls` for the first argument of classmethods. If you don't need the class in the method, use `@staticmethod` instead.

unflxw: Ah, then these should either be in the `Options` object (and use `cls` to instantiate the class) or be modified to use `@staticmethod`. We can change that in a separate PR.

unflxw (approved):

src/appsignal/config.py
tombruijn: Yes, let's make it a static method then. It doesn't need to know the class's state.

PR #101, opened 2023-07-07
Migrate from docopt to argparse (take 2)

orsinium:
This is a copy of #97 but with lintje violations (hopefully) resolved.

Closes https://github.com/appsignal/appsignal-python/issues/84


orsinium:
Nope, doesn't work.

PR #102, opened 2023-07-07
Migrate from docopt to argparse take 5

orsinium:
1. Remove docopt from dependencies.
1. Use argparse to configure CLI.
1. Use cached_property to ensure on type-level that the API key and app name are always provided when we need them.


orsinium:
Finally!

tombruijn (approved):

PR #103, opened 2023-07-07
Push api key validator

jeffkreeftmeijer:
Currently, validating push API keys is done in the install command. The
upcoming diagnose command also needs to validate push API keys, so this
patch moves the code into a separate module that will be used from both
places.

[skip changeset]


luismiramirez:
I moved the module to the CLI namespace.

unflxw (approved):

tombruijn (changes requested):
Build is red.

luismiramirez:
@tombruijn this is working on the diagnose command PR. Closing this one

PR #113, opened 2023-08-01
Warn about the package name change

unflxw:
Part of #111.

I'm opening this draft PR against the `main` branch, but this will be changed later to an `appsignal-beta` branch. That branch will be based on the first `appsignal` release, and it will be used for the last `appsignal-beta` release, being abandoned afterwards.

[skip changeset] (changeset will already be added by #112)

---

This commit will be added on a final release of the `appsignal-beta` package, informing users about the package name change to `appsignal` and prompting them to update their dependency requirements files.

Issue #111, opened 2023-08-01
Rename package from `appsignal-beta` to `appsignal`


unflxw:
### Plan

- Publish a `v0.2` to `appsignal` with the latest changes, including the package name change.
- Manually publish `v0.2` to `appsignal-beta` as well, with the changes above, as well as a change that, on import, prints to standard error, prompting you to change your dependency to the `appsignal` package. This will be the last `appsignal-beta` package published.

This provides a way for users to change the package name in their dependencies, without dragging in any other changes (`v0.2` on `appsignal-beta` -> `v0.2` on `appsignal`)

## To do

appsignal-beta package has been moved to this repo: https://github.com/appsignal/appsignal-python-beta

- [x] Rename package from `appsignal-beta` to `appsignal` on `main` -- https://github.com/appsignal/appsignal-python/pull/112
- [x] Release `appsignal v0.2` (but not `appsignal-beta`)
- [x] Update the docs: https://github.com/appsignal/appsignal-docs/pull/1443
- [x] Update the installation wizard: https://github.com/appsignal/appsignal-server/pull/9916
- [x] Import appsignal_beta library from appsignal library: https://github.com/appsignal/appsignal-python/pull/115
- [x] Manually release appsignal-beta v0.2 package with dependency on appsignal package: https://github.com/appsignal/appsignal-python-beta
- [x] Archive repo: https://github.com/appsignal/appsignal-python-beta

unflxw:
Perhaps we can turn the `appsignal-beta` package into a stub package that installs the `appsignal` package instead 🤔 I talked about this possibility on Slack: https://appsignal.slack.com/archives/C055A2H9NPJ/p1690970202061239

PR #114, opened 2023-08-04
Add stub `appsignal-beta` package

unflxw:
Part of #111. Alternative implementation to #113. If this is accepted, #113 will be closed.

---

Create an empty package that installs the `appsignal` package as a dependency. Copy the README from the `appsignal` package, with an added deprecation warning.

---

Note that this PR is set to merge into a new, empty `appsignal-beta` branch.

To test that this actually works, we can make use of the existing `appsignal` package in TestPyPI. Run `hatch build`, then:

```
pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple dist/appsignal_beta-0.2.0-py3-none-any.whl
```

Then, in a Python REPL, you should be able to `import appsignal`, and check that `appsignal.__about__.__version__` is `"0.0.3"`, which is the last version of the `appsignal` package that we published on TestPyPI.

The build fails because this new empty branch doesn't even have a `semaphore.yml` file. This is fine, I think.


tombruijn:
Moved to new repo instead to keep the two packages separate going forward: https://github.com/appsignal/appsignal-python-beta

PR #117, opened 2023-08-18
Add the statsd_port config option

tombruijn:
Allow users to configure the statsd_port on the appsignal-agent process. This to fix two conflicting servers on the same port.

I had made this before (and checked Python off from the to do list), but never submitted the change 😖

Part of https://github.com/appsignal/appsignal-agent/issues/982

### DIFF
--- a/.changesets/add-statsd_port-config-option.md
+++ b/.changesets/add-statsd_port-config-option.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add the `statsd_port` config option to change the StatsD UDP server port of the appsignal-agent process. By default the port is set to 8125.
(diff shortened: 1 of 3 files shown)


unflxw (approved):
How come it's a string and not a number? 🤔 I think users will expect to pass `statsd_port=1234` to the config, not `statsd_port="1234"`.

tombruijn:
> How come it's a string and not a number? 🤔 I think users will expect to pass `statsd_port=1234` to the config, not `statsd_port="1234"`. Perhaps it should be typed as `str | int | None`, [like `opentelemetry_port`](https://github.com/appsignal/appsignal-python/pull/118/files#diff-0f75d66c5137c4706b732e2a1923d9682400a8922317e0f4abd0653b40bcac1dR50)?

I ran into this with the OpenTelemetry port as well, fixed it there, but not here. Will fix it now.

Issue #27, opened 2023-04-26
Running multiple apps on one machine


unflxw:
Currently, if you run multiple AppSignal for Python apps in the same machine with different configurations, only one agent will be able to report data, as only one agent will be able to listen for OpenTelemetry traces on port 8099.

[They will also step over each other as they will share the same working directory, but this will be fixable once #16 implements `working_dir_path`.]

We discussed this in [this Slack conversation](https://appsignal.slack.com/archives/C055A2H9NPJ/p1682517756326009?thread_ts=1682515629.047589&cid=C055A2H9NPJ). The easy fix is to implement some sort of `opentelemetry_port` config option.

This issue already exists in the Python OpenTelemetry setup, so it's not a blocker for release.

## To do

- [x] Add `opentelemetry_port` config option to:
  - [x] Agent config, see also `statsd_port` config option for an example
  - [x] Python package: https://github.com/appsignal/appsignal-python/pull/118
- [x] Document
  - [x] Config option in Python
  - [x] The need for `opentelemetry_port` and `working_directory_path` config options to be set and how (unique per app).

tombruijn:
Python package 0.2.2 released along with docs updates (currently deploying at time of writing) on how to make this work: https://docs.appsignal.com/guides/application/multiple-applications-on-one-host.html

PR #118, opened 2023-08-18
Add opentelemetry_port config option

tombruijn:
Allow users to configure the OpenTelemetry HTTP server in the agent to listen on another port.

Part https://github.com/appsignal/appsignal-python/issues/27
Requires an agent update before merge and release.

## To do

- [ ] Add agent update to PR

### DIFF
--- a/.changesets/add-opentelemetry-port-config-option.md
+++ b/.changesets/add-opentelemetry-port-config-option.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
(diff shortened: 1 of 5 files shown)


unflxw (changes requested):
The port that the Python OpenTelemetry exporter sends data to also needs to be changed based on this environment variable: https://github.com/appsignal/appsignal-python/blob/main/src/appsignal/opentelemetry.py#L106

tombruijn:
Damn then we do need to set it by default in the integration as well and that requires some more changes.

unflxw (approved):

luismiramirez (approved):

PR #119, opened 2023-08-18
Add host_role config option

tombruijn:
Allow users to configure a role for a host. It's not used a lot in the product yet, but we generate a metric for it and we can do some more stuff with it in the future.

I chose the `host_role` name because `role` alone is a bit ambiguous and the env var is named `APPSIGNAL_HOST_ROLE`.

Part of https://github.com/appsignal/appsignal-agent/issues/1020

### DIFF
--- a/.changesets/add-role-config-option.md
+++ b/.changesets/add-role-config-option.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add the `role` config option. This config option can be set per host to generate some metrics automatically per host and possibly do things like grouping in the future.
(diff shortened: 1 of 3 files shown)


tombruijn:
> I see how it would be preferable for it to be named `host_role`, but `role` is also fine, and it's already in use by [the stand-alone agent](https://github.com/appsignal/appsignal-agent/blob/fac4694a4126b1692ade56c0c11f13de82128def/agent/src/standalone_config.rs#L73).
> 
> (I'm just pointing out the existing mismatch -- I'm happy to keep either name or both)

We can still rename it in the standalone agent! It's not documented anywhere 🤫

unflxw:
Either/both of those sound good to me.

unflxw (approved):

luismiramirez (approved):

tombruijn:
Cherry picked in c6e1da386d0f7a372913a8b7e7c1ba26732eae2c

PR #120, opened 2023-08-25
Use relative imports in the CLI modules

tombruijn:
When using `python -m appsignal ...` it would complain about missing imports.

Use relative imports instead, they seem to work for other imports, like `..__about__` in `src/appsignal/cli/version.py`.


unflxw (approved):
I will always be confused about relative imports.

luismiramirez (approved):

Issue #7, opened 2023-03-24
Implement metrics


unflxw:
Add metrics to the Python integration using OpenTelemetry.

## To do

- [x] Research how to implement metrics
  - [x] OpenTelemetry metrics
    - [x] Add OpenTelemetry metrics support to the agent (include the OpenTelemetry metrics protobuf)
- [x] Convert absolute counters to incremental counters. Now it's stored as a gauge, but we should calculate the delta and send that. -- https://github.com/appsignal/appsignal-agent/pull/1050
- [x] Add metrics exporter to the Python package - https://github.com/appsignal/appsignal-python/pull/121
- [x] Document how to use OpenTelemetry metrics to report custom metrics.
  - Document which OpenTelemetry metric types we convert to which AppSignal metric types.
  - Add note that we don't support histograms.

When this is done we have custom metrics support which is good enough. Close the issue and spin the next steps off into separate issues.

## Next steps
- https://github.com/appsignal/appsignal-agent/issues/1042
- #129
- #130
- #131

unflxw:
Made some progress on this task, [as mentioned on Slack](https://appsignal.slack.com/archives/C055A2H9NPJ/p1687963877754179).

First, a note: the OpenTelemetry docs say that [there is no OTLP/HTTP exporter for metrics](https://opentelemetry.io/docs/instrumentation/python/exporters/#using-http), but this isn't actually true. [The exporter was implemented months ago](https://github.com/open-telemetry/opentelemetry-python/pull/2891) and it's functional, but it's not documented.

Once that was figured out, I managed to get some metrics sent from the Python integration to the agent, and parsed on the agent’s side. Here’s the debug output from the agent: https://gist.github.com/unflxw/d34dd4c37a72269d5b1e1615a53b4411

Here’s the changes [in the Python integration](https://github.com/appsignal/appsignal-python/compare/add-opentelemetry-metrics?expand=1) and [in the Flask test setup](https://github.com/appsignal/test-setups/compare/add-flask-custom-metrics-endpoint?expand=1), and here’s the (extremely work-in-progress!) changes [in the agent](https://github.com/appsignal/appsignal-agent/compare/wip-opentelemetry-metrics?expand=1).

The next step will be to make the agent turn these into AppSignal metrics. Some, such as counters, have a clear mapping to existing metrics. Others, such as histograms, not so much. 

[...]

PR #121, opened 2023-08-25
Add OpenTelemetry metrics exporter

tombruijn:
## Add OpenTelemetry metrics exporter

Add the metrics exporter by default so users are able to send metrics using OpenTelemetry to AppSignal.

Tracking issue: https://github.com/appsignal/appsignal-python/issues/7

## Configure the metrics export preferred_temporality

We can set with which temporality metrics are exported. To avoid clashes with cumulative sum conversion to deltas from multiple processes, prefer to send cumulative sums as delta sums.

To make it explicit for all metrics we support, I've added them all the exporter config.

### DIFF
--- a/.changesets/enable-opentelemetry-metrics-exporter.md
+++ b/.changesets/enable-opentelemetry-metrics-exporter.md
@@ -1,0 +1,6 @@
+---
+bump: "minor"
+type: "add"
+---
+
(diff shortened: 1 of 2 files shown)


tombruijn:
Agent update merged in #123 because it also included other fixes. This PR is ready to merge.

luismiramirez (approved):

unflxw (approved):

Issue #124, opened 2023-09-05
Metric helpers


unflxw:
Implement helpers to easily send custom metrics to AppSignal without having to understand OpenTelemetry metrics.

Ideally expose the same API as the Ruby implementation: https://docs.appsignal.com/metrics/custom.html#gauge

Note that we currently only support OpenTelemetry counters and gauges, not OpenTelemetry histograms (which would be AppSignal distribution values)

See #122 for sample data helpers. Will be documented as part of https://github.com/appsignal/appsignal-docs/issues/1492.

tombruijn:
Also related to the metrics tracking issue https://github.com/appsignal/appsignal-python/issues/7

tombruijn:
[Related Slack discussion](https://appsignal.slack.com/archives/C055A2H9NPJ/p1693919028620929) about how to make gauges work that have a somewhat similar API to our other AppSignal integrations.

PR #127, opened 2023-09-06
Add metric helpers for counters and gauges

tombruijn:
## Add counter helper for reporting counter metrics

Add convenience helper method to report counter metrics using the AppSignal terminology along with all the OpenTelemetry boilerplate.

This doesn't allow for setting units or metric descriptions, which OpenTelemetry supports, because we discard this metadata.

Part of #124

## Add gauge helper for reporting gauge metrics

Add convenience helper method to report gauge metrics using the AppSignal terminology along with all the OpenTelemetry boilerplate.

This doesn't allow for setting units or metric descriptions, which OpenTelemetry supports, because we discard this metadata.

This is using the OpenTelemetry ObservableGauge in the background, but exposes an API to what we're used to in other integrations with `set_gauge`.

The ObservableGauge registers a callback that it runs asynchronously in the background every 60 seconds. 60 seconds after the app starts it begins and when the app exits it's called one more time.

This makes the API simpler for our users so they don't have to worry about setting up methods for callbacks and such. It's more limiting, but people can always use the OpenTelemetry metrics module themselves to send metrics.

More information about the OpenTelemetry metrics SDK can be found here: https://opentelemetry-python.readthedocs.io/en/latest/sdk/metrics.html

Part of #124

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -1,0 +1,30 @@
+from opentelemetry.metrics import CallbackOptions, UpDownCounter
+
+from appsignal.metrics import _counters, increment_counter
+
+
+def test_increment_counter_creates_new_counter():
+    assert _counters.get("metric_name") is None
+
+    increment_counter("metric_name", 1)
+
+    assert isinstance(_counters["metric_name"], UpDownCounter)
+
+
+def test_increment_counter_creates_updates_existing_counter(mocker):
+    increment_counter("metric_name", 1)
+
+    counter_spy = mocker.spy(_counters["metric_name"], "add")
+    increment_counter("metric_name", 1)
+
+    counter_spy.assert_called_once_with(1, None)
+
+
+def test_increment_counter_with_tags(mocker):
+    increment_counter("metric_name", 1)
+
+    counter_spy = mocker.spy(_counters["metric_name"], "add")
+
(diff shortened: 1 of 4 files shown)


tests/test_metrics.py
```diff
@@ -0,0 +1,95 @@
...
+    meter_spy.assert_called_once_with("metric_name", callbacks=mocker.ANY)
+    callbacks = meter_spy.call_args.kwargs["callbacks"]
+    assert len(callbacks) == 1
+
+    # Mock the ObservableGauge's async task calling the callbacks
+    return_values = callbacks[0](CallbackOptions())
+    assert len(return_values) == 1
+    return_value = return_values[0]
+    assert return_value.value == 10
+    assert return_value.attributes is None
```
unflxw, line 56: I think it would be good to assert that, after the gauge callback has been called, the corresponding entry in the gauge value dictionary is empty (`_gauges["metric_name"] == {}`) and/or that calling the callback again results in no observations being sent.

(Same for the `with_tags` scenario)

unflxw (approved):
Looks great! Left a comment on testing what happens when `set_gauge` has not been called between callback invocations.

tests/test_metrics.py
tombruijn: Good catch! Amended extra assertions

luismiramirez (approved):

PR #128, opened 2023-09-06
Sample data helpers

### LINKED ISSUE #122: Sample data helpers
Implement sample data helpers, [like those implemented for Node.js](https://docs.appsignal.com/guides/custom-data/sample-data.html#nodejs).

unflxw:
Closes #122. This follows the design and implementation of the Node.js and Ruby helpers of the same name, except for `send_error`, [as discussed in Slack](https://appsignal.slack.com/archives/C055A2H9NPJ/p1694006687516139) and described in the commit message.

Long story short, unlike with Ruby blocks, a context manager **must** be used with a `with` block, and accidentally calling it without a `with` block will silently do nothing. So it's been split into two functions: `send_error`, which does _not_ take a `with` block and sends the error without any additional context, and `send_error_with_context`, which takes a `with` block where the user can add additional context.

---

### [Implement sample data helpers](https://github.com/appsignal/appsignal-python/commit/dc6985aabbcac40bde0ae417766282683051cf85)

Implements sample data helpers that simplify adding AppSignal magic
attributes to the current OpenTelemetry span.

### [Add error helpers](https://github.com/appsignal/appsignal-python/commit/ebabaac86be956333cc8a22555012f5205f69be1)

The `set_error` helper follows the implementation of the Node.js
integration.

For the `send_error` helper, it has been split into two: a "naked"
`send_error` function, that just sends the error without added span
context, and a `send_error_with_context` function, which must be
used with a `with` block to which the span is provided as an
argument.

---

I think I'll have another go at using the OpenTelemetry MemoryExporter somehow, instead of mocking so much stuff. For the sample data helpers it's okay, but for the error helpers the result is an incomprehensible mess that doesn't meaningfully test anything. Seven `mocker.Mock()` invocations in the same test case is at least five too many, I think.


src/appsignal/__init__.py
```diff
@@ -1,7 +1,32 @@
...
+    "set_category",
+    "set_custom_data",
+    "set_header",
+    "set_name",
+    "set_namespace",
+    "set_params",
+    "set_root_name",
+    "set_session_data",
+    "set_tag",
+]
```
tombruijn: Oh we can export things on the "main" module.

Should we do the same for the metrics helpers in #127 or should this use that PR's approach of having to import using `appsignal.tracing`?

src/appsignal/helpers.py
tombruijn, line 1: To help distinguish the helpers from the metric helpers, shall we rename this to `tracing.py`?

src/appsignal/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+
+def set_attribute(attribute: str, value: Any, span: Span | None = None) -> None:
+    span = span or trace.get_current_span()
+    # TODO: `trace.get_current_span` will return an INVALID_SPAN
+    # if there is no current span. Should we log something about
+    # it? The Node.js integration does log something about it.
+    #
+    # To detect if it's an invalid span, it seems we would have
+    # to do something like:
+    # `span.get_span_context().is_valid`
```
tombruijn, line 26: Would be helpful to log something about it 👍

src/appsignal/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+    span.set_attribute(attribute, value)
+
+
+def set_serialised_attribute(
```
tombruijn, line 30: Shall we make this private with an underscore prefix?

```suggestion
def _set_serialised_attribute(
```

src/appsignal/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+    attribute: str, value: Any, span: Span | None = None
+) -> None:
+    serialised_value = json.dumps(value)
+    set_attribute(attribute, serialised_value, span)
+
+
+def set_prefixed_attribute(
```
tombruijn, line 37: Shall we make this private with an underscore prefix?

```suggestion
def _set_prefixed_attribute(
```

src/appsignal/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+    # To detect if it's an invalid span, it seems we would have
+    # to do something like:
+    # `span.get_span_context().is_valid`
+    span.set_attribute(attribute, value)
+
+
+def set_serialised_attribute(
+    attribute: str, value: Any, span: Span | None = None
+) -> None:
+    serialised_value = json.dumps(value)
```
tombruijn, line 33: You say `Any` is allowed here. Does the JSON dump never fail, even when it's given a Python class or something?

src/appsignal/helpers.py
```diff
@@ -77,3 +79,27 @@ def set_namespace(namespace: str, span: Span | None = None) -> None:
...
+
+
+@contextmanager
+def send_error_with_context(error: Exception) -> Iterator[Span]:
+    name = error.__class__.__name__
+
+    # TODO: this needs to always be a root span somehow!!!!
+    with _send_error_tracer.start_as_current_span(name) as span:
+        set_error(error, span)
+        yield span
```
tombruijn, line 105: We can also leave this helper out until we have a way that it behaves like it should: setting the error on a new root span.

.changesets/add-sample-data-and-error-helpers.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add sample data and error helpers.
```
tombruijn, line 6: Some usage examples (that we can later copy-paste) in the docs would be helpful. See PR #124 for examples.

tombruijn (commented):
Some thoughts for the review so far.

tombruijn:
> I think I'll have another go at using the OpenTelemetry MemoryExporter somehow, instead of mocking so much stuff. For the sample data helpers it's okay, but for the error helpers the result is an incomprehensible mess that doesn't meaningfully test anything. Seven `mocker.Mock()` invocations in the same test case is at least five too many, I think.

Please do! :)

src/appsignal/helpers.py
unflxw: Oh, yes, you're right. The dump can fail when the class is not serializable:

```python
class Foo:
    pass

json.dumps(Foo())
# TypeError: Object of type Foo is not JSON serializable
```

And also when the object is circular, which annoyingly, unlike in Node.js, I couldn't find a way to fix:
```
x = {}
x["x"] = x

json.dumps(x)
# ValueError: Circular reference detected
```

Not sure if this can be expressed on the type system, though. I'll have a look. But regardless, we should still account for it at run-time, and probably log a warning instead of letting the error bubble through.

src/appsignal/helpers.py
unflxw: Yes! 👍  Same for `set_attribute` and `set_prefixed_attribute`.

src/appsignal/__init__.py
unflxw: I'm happy either way! I'm exposing them on the main module because this is what we do on Ruby and Node.js, where it's `Appsignal.set_error` and `import { setError } from "@appsignal/nodejs"`. But we don't need to do that.

Whatever we choose to do, let's do the same for these helpers and for the metrics helpers.

src/appsignal/helpers.py
unflxw: Fixed.

unflxw:
@tombruijn Done! Much better now, I think.

You might want to use it for the metrics tests as well! I left that ready on https://github.com/appsignal/appsignal-python/pull/128/commits/a7d90e5cd9bea5b3651426bca96be4882e7abbb2.

.changesets/add-sample-data-helpers.md
```diff
@@ -0,0 +1,17 @@
...
+```python
+from appsignal import set_custom_data
+
+# Must be used in an instrumented context -- e.g. a Flask handler
+@app.get("/")
+def index():
+  set_custom_data({"hello": "there"})
+```
+
+The full list of sample data helpers is available in our documentation.
```
tombruijn, line 17:
```suggestion
The full list of sample data helpers is available in our documentation as part of the [tagging and sample data guides](https://docs.appsignal.com/guides/custom-data.html) and [namespace guide](https://docs.appsignal.com/guides/namespaces.html).
```

I suggest we link to the docs where most of them will be described once https://github.com/appsignal/appsignal-docs/issues/1492 is done.

.changesets/add-error-helpers.md
```diff
@@ -0,0 +1,17 @@
...
+```python
+from appsignal import send_error
+
+def index():
+  cookies = put_hand_in_cookie_jar()
+  except HandStuckError as error:
+    send_error(error)
+```
+
+More information about error helpers is available in our documentation.
```
tombruijn, line 17:
```suggestion
More information about [error helpers](https://docs.appsignal.com/python/custom-instrumentation/exception-handling.html) is available in our documentation.
```

I suggest we link to this location. It doesn't exist yet, but we have a similar page for other integrations, so we should make one for Python too.

tombruijn (approved):

src/appsignal/__init__.py
tombruijn: Let's export it on the `__all__` thing so it's all on the same import. I assume that if people find the collection of helpers annoying to import, they can also import `appsignal.tracing` and call `tracing.set_...` like it's done now for the metrics helpers.

I'll also update the metric helpers to export them this way, but after your PR is merged.

tombruijn:
> You might want to use it for the metrics tests in #127 as well! I left that ready on a separate commit on [a7d90e5](https://github.com/appsignal/appsignal-python/commit/a7d90e5cd9bea5b3651426bca96be4882e7abbb2), so you can just cherry-pick that in your PR. (Though, it's not entirely clear to me how you'd use it with observables...)

Great! I'll have a look, but let's not block merging #127 on it.

tombruijn:
@unflxw can you create a PR for the Python test apps to use these helpers?

luismiramirez (approved):

PR #135, opened 2023-10-02
Update tag value types to support int, float, bool

tombruijn:
We support these other types as tag values, so let's update the type definition to allow it.

[skip changeset] as it's an improvement on a not yet released feature.


tombruijn:
Merging because I want to release this with the new package version.

Issue #138, opened 2023-10-16
Demo CLI should have --environment option


tombruijn:
Allow people to specify the app environment via the CLI. It now defaults to `development` which is not always what people want. And we don't want people to have to use the env var either.

- [ ] Add `--environment` CLI config option to diagnose command.
- [x] Add `--environment` CLI config option to demo command.
- [x] Update demo CLI docs: https://docs.appsignal.com/python/command-line/demo.html
- [x] Update docs examples: https://docs.appsignal.com/support/debugging.html

tombruijn:
The diagnose command doesn't have any of the `--application` or `--push-api-key` options, so maybe it's not a good idea to add `--environment` either. Instead load the config file in #142 to load the config.

tombruijn:
I think there's less of a need to add the `environment` option to the diagnose CLI. Let's revisit it when we run into it. Closing.

unflxw:
@tombruijn Agreed. In Ruby, it makes sense to have the `--environment` option, as the configuration that is loaded from `appsignal.yml` can be different for each environment. That's not how it works here.

PR #146, opened 2023-10-18
Add demo CLI test and remove validate output

tombruijn:
Add the missing test for the demo CLI. It only tests if the config is valid or not, as testing if the OpenTelemetry span creation works is not necessary I think for now. It will require a lot of mocking, which doesn't tell us if the code actually still works or not.

I've also updated the config validation to not print an error immediately on validation. This may also be printed in some scenarios we don't want.

Preferably this is logged to the `appsignal.log` file like in the Ruby integration, but here we don't have a logger instance yet in the Python package. Something to improve later.

The demo CLI will log that the config is invalid, which can be debugged with the diagnose CLI if needed. The normal client start already logged a message when it couldn't start.

Needed for #138

### DIFF
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -300,8 +300,6 @@ class Config:
         push_api_key = self.option("push_api_key") or ""
         if len(push_api_key.strip()) > 0:
             self.valid = True
-        else:
-            print("appsignal: Push API key not set after loading config. Not starting.")
 
 
 def parse_bool(value: str | None) -> bool | None:
--- a/.changesets/validate-config-before-starting.md
+++ b/.changesets/validate-config-before-starting.md
@@ -3,4 +3,4 @@ bump: "patch"
 type: "fix"
 ---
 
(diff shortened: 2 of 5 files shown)


src/appsignal/config.py
```diff
@@ -300,8 +300,6 @@ def _validate(self) -> None:
         push_api_key = self.option("push_api_key") or ""
         if len(push_api_key.strip()) > 0:
             self.valid = True
-        else:
-            print("appsignal: Push API key not set after loading config. Not starting.")
```
unflxw, line 304: Hm, not sure about this. I think it's useful to inform the user if this happens at their application's run-time. (And I think it's preferable to have a redundant message be emitted on the demo, than it is to not emit it on run-time at all)

unflxw (commented):
✅ on the tests; see comment

src/appsignal/config.py
tombruijn: I think it's useful to add, and I copied it from the Ruby gem that does it, but it adds it to its logger, not always print it.

Other integrations don't do this. So until we have a better way to log it, I want to remove it.

PR #147, opened 2023-10-18
Add demo CLI --environment option

tombruijn:
## Move empty line print from demo CLI to install CLI

Move the empty line print to the install CLI so that the demo CLI doesn't start with an empty line. Update the tests to match.

## Define CLI options in each CLI class

Instead of defining some defaults in the main CLI class, explicitly define them per CLI command, so it's easier to customize the CLI options per command.

## Add demo CLI --environment option

Allow users to specify which environment they want to send the demo samples for. Normally it defaults to the `development` environment.

For the install CLI (which calls the demo CLI), it will continue to default to the development environment like our other integrations.

Part of #138

---

Based on #146

### DIFF
--- a/tests/cli/test_demo.py
+++ b/tests/cli/test_demo.py
@@ -23,7 +23,7 @@ def test_demo_with_valid_config(mocker, capfd):
         assert main(["demo"]) == 0  # Successful run
 
     out, err = capfd.readouterr()
-    assert out.find("Sending example data to AppSignal...") > 0
+    assert out.find("Sending example data to AppSignal...") > -1
 
 
 def test_demo_with_invalid_config(mocker, capfd):
@@ -35,4 +35,4 @@ def test_demo_with_invalid_config(mocker, capfd):
         assert main(["demo"]) == 1  # Exit with error
 
     out, err = capfd.readouterr()
-    assert out.strip() == "AppSignal not starting: no active config found."
+    assert out.find("AppSignal not starting: no active config found.") > -1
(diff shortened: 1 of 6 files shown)


tests/cli/test_demo.py
```diff
@@ -23,7 +23,7 @@ def test_demo_with_valid_config(mocker, capfd):
         assert main(["demo"]) == 0  # Successful run
 
     out, err = capfd.readouterr()
-    assert out.find("Sending example data to AppSignal...") > 0
+    assert out.find("Sending example data to AppSignal...") > -1
```
luismiramirez, line 30: Isn't `>=` a valid operator in Python?

tombruijn: I think you left this comment on an older comment and I already fixed it in a new comment.

luismiramirez: Do you have a link? That completely disappeared from my head

tombruijn: Oh no I got confused with a newer branch that isn't created yet. But I think this is fine. `-1` indicates it couldn't be found. `0` means it's the first character in the string. So if it's larger than -1 it could be found.
`0` worked before because the demo CLI output started with an empty line that I [(re)moved in this change](https://github.com/appsignal/appsignal-python/pull/147/commits/d6089bde712760dc5714b9929d39c2649362d839).

luismiramirez: My suggestion was to use `>= 0` instead, but it's a personal preference. I'm also good with the `-1`

unflxw (approved):

Issue #142, opened 2023-10-18
Diagnose CLI should load __appsignal__.py if present


tombruijn:
To provide a complete picture in the diagnose report, the diagnose CLI should load the `__appsignal__.py` config file. Without it, we have an incomplete configuration and can't properly diagnose it with all the information.

## To do

- [x] Load the `__appsignal__.py` file in the **diagnose CLI** when it's found and make sure it's config is part of the diagnose report.
- [x] Load the `__appsignal__.py` file in the **demo CLI** when it's found.

tombruijn:
This might also be useful for the demo CLI.

PR #149, opened 2023-10-18
Load __appsignal__.py file when found in demo CLI

tombruijn:
## Extract demo samples creation to Demo class

Create the demo samples in its own class, separate from the configuration. This makes it more easily called by other parts of the package, like the install CLI.

The install CLI will configure the client on its own, as it uses different arguments than the demo CLI. Passing the CLI arguments along will cause an error if the demo if it calls demo CLI specific arguments that aren't part of the install CLI.

## Load __appsignal__.py file when found in demo CLI

When the demo CLI finds a `__appsignal__.py` configuration file, use that for the configuration, rather than only rely on the CLI prompt.

The demo CLI will overwrite the `active` config option to always be true when loading the file config.

One install test failed, because it expects a real file to be present and I didn't want to add more mocking to account for that. Instead, it now doesn't mock and uses a dummy project on the file system.

Part of #142

### DIFF
--- a/src/appsignal/cli/demo.py
+++ b/src/appsignal/cli/demo.py
@@ -39,6 +39,14 @@ class DemoCommand(AppsignalCLICommand):
         print(f"Starting AppSignal client for {self._name}...")
         client.start()
 
+        Demo.transmit()
+
+        return 0
+
+
+class Demo:
+    @staticmethod
+    def transmit() -> None:
         tracer = trace.get_tracer(__name__)
 
         # Performance sample
@@ -66,5 +74,3 @@ class DemoCommand(AppsignalCLICommand):
                 raise ValueError("Something went wrong")
             except ValueError as e:
                 span.record_exception(e)
-
-        return 0
--- a/src/appsignal/cli/demo.py
+++ b/src/appsignal/cli/demo.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import os
 from argparse import ArgumentParser
+from runpy import run_path
 
 from opentelemetry import trace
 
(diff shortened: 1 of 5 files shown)


src/appsignal/cli/demo.py
```diff
@@ -19,26 +22,73 @@ def init_parser(parser: ArgumentParser) -> None:
...
+            if not spec:
+                # Not sure when this happens, but mypy complains that `spec`
+                # can be Optional.
+                print("Invalid __appsignal__.py config file found. Exiting.")
+                return 1
+
+            app_config_module = module_from_spec(spec)
+            loader.exec_module(app_config_module)
+            try:
+                client = app_config_module.appsignal
```
unflxw, line 39: Nit: I don't think there's anything wrong with this implementation, but an easier way to do this might be using [the `runpy` module](https://docs.python.org/3/library/runpy.html), which _I think_ uses the same mechanisms internally, but exposes them in a less piece-meal manner:

```python
from runpy import run_path
client = run_path(app_config_path)["appsignal"]
```

We already use this [in our Hatch build hooks](https://github.com/appsignal/appsignal-python/blob/main/src/scripts/build_hook.py#L17-L22), so there would be an advantage in not having to grok another way of doing the same thing.

tests/cli/test_demo.py
```diff
@@ -61,3 +64,87 @@ def test_demo_with_invalid_config(mocker, capfd):
...
+
+    # Change back to original dir
+    os.chdir(request.config.invocation_params.dir)
+
+    out, err = capfd.readouterr()
+    assert out.find("Sending example data to AppSignal...") > -1
+    assert out.find("Starting AppSignal client for CLI app...") > -1
+
+
+def test_demo_with_invald_config_file(request, mocker, capfd):
```
unflxw, line 119:
```suggestion
def test_demo_with_invalid_config_file(request, mocker, capfd):
```

tests/cli/test_install.py
```diff
@@ -41,6 +44,15 @@ def assert_wrote_file_contents(mocker):
     )
 
 
+def assert_wrote_real_file_contents(test_dir, name, push_api_key):
```
unflxw, line 47: Do we want to keep both this and `assert_wrote_file_contents`? (the mocker-based version, above)

(OK with keeping it, just making sure it's intentional)

unflxw (approved):

src/appsignal/cli/demo.py
tombruijn: That's interesting. And runpy is just built in. Why couldn't I find a good resource for it then.

tests/cli/test_install.py
tombruijn: The other one is still used by other specs. I'd be okay with rewriting them all, but not doing that in this PR.

PR #163, opened 2023-10-27
Fix push API key validation on installer

### LINKED ISSUE #161: Installer API key validation always fails
Steps to reproduce:
- Click "New app" in the UI for your AppSignal organization's UI
- Run the installation command: `python -m appsignal install --push-api-key xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` (enter any name when prompted)
Expected behaviour: the installation takes place
Actual behaviour: "API key xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx is not valid; please get a new one on https://appsignal.com"

unflxw:
The configuration obtained from the installer prompts was not being passed correctly to the installer.

Fixes #161.

### DIFF
--- a/.changesets/fix-push-api-key-validation.md
+++ b/.changesets/fix-push-api-key-validation.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix an issue where the installer would always find the push API key to be invalid, halting the installation process.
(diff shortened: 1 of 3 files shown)


tombruijn (approved):
Let's merge this and get it shipped. I'd also like to include #159 in that next release.

Some thoughts we can improve upon later: I want a better test for the install CLI behavior, because this should have been caught by the test suite as well. Like an assertion what kind of query params the validation request includes.

PR #165, opened 2023-10-30
Fix the demo samples recognition and use a custom DemoError class

tombruijn:
## Fix the demo samples recognition in the front-end

To display the demo samples message for the first error reported by our integration, fix how the `demo_sample` tag is set.

It should be set as a tag, using the magic attribute, not by directly setting an OpenTelemetry span attribute.

## Use custom DemoError class for Demo error sample

We reported the demo error as a ValueError, which is a normal error in Python. This might have people think something is wrong immediately. Use the custom `DemoError` class to communicate that it's just an example error. This matches how we report this error name in other integrations.

### DIFF
--- a/src/appsignal/cli/demo.py
+++ b/src/appsignal/cli/demo.py
@@ -5,7 +5,7 @@ from argparse import ArgumentParser
 from opentelemetry import trace
 
 from ..client import Client, InvalidClientFileError
-from ..tracing import set_params
+from ..tracing import set_params, set_tag
 from .command import AppsignalCLICommand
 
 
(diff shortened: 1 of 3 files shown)


src/appsignal/cli/demo.py
```diff
@@ -78,20 +78,18 @@ def transmit() -> None:
...
-                "demo_sample",
-                True,
-            )
+            set_tag("demo_sample", True, span)
             try:
-                raise ValueError("Something went wrong")
-            except ValueError as e:
+                raise DemoError("Something went wrong")
+            except DemoError as e:
                 span.record_exception(e)
```
unflxw, line 91: nit: in the line of the other changes to use `set_tag` and `set_params`, this could be `set_error(e)`

unflxw (approved):

src/appsignal/cli/demo.py
tombruijn: I've changed it in #166

Issue #172, opened 2023-11-01
Report in diagnose report if app is running in a container or not


tombruijn:
We currently have the Host sub report's `running_in_container` field for most integration's diagnose reports. But this isn't sent in the Python report.

Probably because we fetch this from the extension normally, but we don't have an extension here. We should add it to the agent's diagnose report instead and merge it into the Host sub report.

PR #173, opened 2023-11-01
Report running_in_container in diagnose report

tombruijn:
I've updated the agent to report the `running_in_container` field in the agent's diagnose report's `host` sub report.

In the Python package I've copied it to the host sub report on the main diagnose report as well. That way the diagnose viewer has to less about where this field is originates from for every integration. It will report this field for the agent report, as for all other integrations, but that won't get in the way. We can even hide it in the viewer.

Part of https://github.com/appsignal/appsignal-python/issues/172

⚠️ Requires agent update in https://github.com/appsignal/appsignal-agent/pull/1073 to make the build pass.

### DIFF
--- a/.changesets/bump-agent-to-b604345.md
+++ b/.changesets/bump-agent-to-b604345.md
@@ -1,0 +1,12 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Bump agent to b604345.
+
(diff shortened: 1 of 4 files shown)


unflxw (approved):

unflxw:
Could the agent update be added as part of this PR before merging?

luismiramirez (approved):

PR #176, opened 2023-11-21
Fix missing platform file error

unflxw:
Intercom conversation: https://app.intercom.com/a/inbox/yzor8gyw/inbox/shared/all/conversation/16410700274092#part_id=comment-16410700274092-21909541622

The platform file should never be missing. But if it somehow is, then don't crash the diagnose report because of it.

Fix a typo in the name of the function that reads from the platform file.

If the platform file is missing from our agent build cache, make sure that we do not consider that to be a correct build. Downloading the agent again is overkill, but it does the trick.

Ensure the platform file is actually added by declaring it as a build artifact.

### DIFF
--- a/src/appsignal/agent.py
+++ b/src/appsignal/agent.py
@@ -41,9 +41,12 @@ class Agent:
             [self.agent_path, "--version"], capture_output=True
         ).stdout.split()[-1]
 
-    def architecutre_and_platform(self) -> list[str]:
-        with open(self.platform_path) as file:
-            return file.read().split("-", 1)
+    def architecture_and_platform(self) -> list[str]:
+        try:
+            with open(self.platform_path) as file:
+                return file.read().split("-", 1)
+        except FileNotFoundError:
+            return ["", ""]
 
 
 agent = Agent()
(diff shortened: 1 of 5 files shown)


luismiramirez (approved):

unflxw:
Tip: you can see which files are included in a `.whl` by running `tar -ztvf dist/whatever.whl` -- they're just tars!

src/appsignal/agent.py
```diff
@@ -41,9 +41,12 @@ def version(self) -> bytes:
...
 
-    def architecutre_and_platform(self) -> list[str]:
-        with open(self.platform_path) as file:
-            return file.read().split("-", 1)
+    def architecture_and_platform(self) -> list[str]:
+        try:
+            with open(self.platform_path) as file:
+                return file.read().split("-", 1)
+        except FileNotFoundError:
+            return ["unknown", "unknown"]
```
tombruijn, line 49: I think two empty strings would be fine here. That would trigger validations in the viewer about missing field values.

```suggestion
            return ["", ""]
```

tombruijn (approved):

PR #187, opened 2024-01-11
Use GitHub Actions as the CI

tombruijn:
Move away from Semaphore and use GitHub Actions as our CI for this project.

Semaphore doesn't have Python 12 yet for us to test against, it will update versions once a month, and with GitHub Actions we can run it on the CI today.

[skip changeset]


tombruijn:
This will need some more work, like auto cancel builds when you push a commit on a branch when another build for an older commit is already running and scheduled builds so we can see if anything fails (and connect that to Slack if possible). But this PR is here to gauge our interest.

unflxw:
I am certainly interested! I think this can replace our Semaphore usage as-is.

Before merging this we should actually see the checks fail as well as succeed. I don't trust it being green unless I have also seen it being red.

unflxw (approved):
See the other comment re: seeing the build fail.

After merging this we'll have to disable Semaphore for this repo so it doesn't mark each build as red.

tombruijn:
@unflxw here's an example with a failure: https://github.com/appsignal/appsignal-python/actions/runs/7492027115

PR #199, opened 2024-03-14
Implement minutely probes

### LINKED ISSUE #129: Add minutely probes system
Blocked by https://github.com/appsignal/appsignal-agent/issues/1042

Add similar system like we have for other integrations: https://docs.appsignal.com/ruby/instrumentation/minutely-probes.html

unflxw:
### [Implement minutely probes](https://github.com/appsignal/appsignal-python/pull/199/commits/0f484200ec174e7bccfc513fc0889b5f627a9484)

Implement a minutely probes system, heavily borrowing from the one
in our Ruby integration.

### [Implement enable_minutely_probes config option](https://github.com/appsignal/appsignal-python/pull/199/commits/521d97daffdb7d8e0cd0c3b03bd1e08e7826bfa3)

Implement a config option that globally enables or disables the
minutely probes system, in line with our other integrations.

When disabled by the config option, the minutely probes system can
still be manually enabled by calling `appsignal.probes.start()`.

Ensure that calling this function multiple times will not result in
the probes running multiple times per minute.

Add a lock around registering, unregistering and executing probes
to prevent race conditions where e.g. a probe's unregistering takes
place halfway through its execution.

### [Rename _opentelemetry functions](https://github.com/appsignal/appsignal-python/pull/199/commits/47f2a8d07e822ef65f3db6d40b1328bdd12742c0)

These functions are already in the `appsignal.opentelemetry` method,
they do not need to have `_opentelemetry` in their name.

### [Test start and stop methods](https://github.com/appsignal/appsignal-python/pull/199/commits/1e348c951e7420032bf9158ee28a91c2e69e17ef)

Got in a bit of a pickle to make the thread initialization testable.
When `client.start()` was called, a thread was spawned, so I had to
implement a way to stop these threads from executing during other
tests (hence `client.stop()`)

All other tests continue to call `_run_probes` directly because the
resulting behaviour is easier to reason about.

Fixes #129. See https://github.com/appsignal/test-setups/pull/189 for a test setup.

### DIFF
--- a/.changesets/add-minutely-probes-system.md
+++ b/.changesets/add-minutely-probes-system.md
@@ -1,0 +1,20 @@
+---
+bump: "minor"
+type: "add"
+---
+
+Add a minutely probes system. This can be used, alongside our metric helpers, to report metrics to AppSignal once per minute.
+
+```python
+from appsignal import probes, set_gauge
+
+def new_carts(previous_carts=None):
+    current_carts = Cart.objects.all().count()
+
+    if previous_carts is not None:
+      set_gauge("new_carts", current_carts - previous_carts)
+
+    return current_carts
+
+probes.register("new_carts", new_carts)
+```
--- a/.changesets/add-minutely-probes-system.md
+++ b/.changesets/add-minutely-probes-system.md
@@ -18,3 +18,5 @@ def new_carts(previous_carts=None):
 
 probes.register("new_carts", new_carts)
 ```
+
+The minutely probes system starts by default, but no probes are automatically registered. You can use the `enable_minutely_probes` configuration option to disable it.
--- a/conftest.py
+++ b/conftest.py
@@ -21,8 +21,8 @@ from appsignal.opentelemetry import METRICS_PREFERRED_TEMPORALITY
 
 @pytest.fixture(scope="function", autouse=True)
 def disable_start_opentelemetry(mocker):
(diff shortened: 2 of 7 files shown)


.changesets/add-minutely-probes-system.md
```diff
@@ -0,0 +1,20 @@
...
+
+def new_carts(previous_carts=None):
+    current_carts = Cart.objects.all().count()
+
+    if previous_carts is not None:
+      set_gauge("new_carts", current_carts - previous_carts)
+
+    return current_carts
+
+probes.register("new_carts", new_carts)
```
unflxw, line 19: @noccor I borrowed from your draft for this example :)

The probe taking an argument and returning a value is optional, the following is also okay:

```python
def new_carts():
  set_gauge("new_carts", ...)

probes.register("new_carts", new_carts)
```

noccor: @unflxw - amazing! Thank you so much for getting this into action so quickly 🙇

jeffkreeftmeijer (approved):

unflxw:
Apologies to @jeffkreeftmeijer, who had already reviewed this, but I ended up significantly altering the implementation to make it thread-safe and testable, so I'm asking for a re-review. 🙇

luismiramirez (approved):

PR #201, opened 2024-03-18
Modify installation instructions to use `import appsignal`

unflxw:
This is a sequel to #200 that I, uh, kinda forgot to do.

---

### [Modify installation instructions](https://github.com/appsignal/appsignal-python/commit/256b45578dd9b68e7f9a992fb9ae42a56635e799)

Add a warning emoji whenever manual instrumentation is required,
that is, always.

For Django's instructions, refer to the modification needed as a
generic "your entrypoints" and leave it for the docs to explain
further.

Change other instructions to use `import appsignal`, replacing the
use of `from __appsignal__ import appsignal`.

### [Specify requests dependency in hook](https://github.com/appsignal/appsignal-python/commit/7cf30c7354542362fac2959c1f67a85716782cad)

I am not sure exactly why this change is needed now. I think Hatch
[added support for specifying custom build hook dependencies][oops]
directly in the custom build hook in the last release, and this
might have accidentally broken the way in which we used to specify
this dependency.

It might also be that the breakage is intentional, in which case
they [forgot to update the docs][whoopsie].

[oops]: https://github.com/pypa/hatch/commit/c04877c183e850200231ed1501033a21db044d30
[whoopsie]: https://hatch.pypa.io/1.9/config/build/#dependencies_1

### DIFF
--- a/src/appsignal/cli/install.py
+++ b/src/appsignal/cli/install.py
@@ -23,6 +23,8 @@ appsignal = Appsignal(
 
 INSTALL_FILE_NAME = "__appsignal__.py"
 
+WARNING_EMOJI = "\u26A0\ufe0f"
+
 
 class InstallCommand(AppsignalCLICommand):
     """Generate Appsignal client integration code."""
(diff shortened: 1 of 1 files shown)


luismiramirez (approved):

src/appsignal/cli/install.py
```diff
@@ -23,6 +23,8 @@
 
 INSTALL_FILE_NAME = "__appsignal__.py"
 
+WARNING_EMOJI = "\u26A0\ufe0f"
```
luismiramirez, line 26: Could this be just the plain emoji?

jeffkreeftmeijer (approved):

src/appsignal/cli/install.py
unflxw: I gave that another try but VSCode insists on saving the non-emoji variant (⚠ not ⚠️) to the file. :(

PR #203, opened 2024-04-09
Implement heartbeats for Python

unflxw:
### [Store global client and expose its config](https://github.com/appsignal/appsignal-python/commit/478c44e2f54360eaadcc8435cd75cdba87eb26f9)

Store the last (hopefully only) initialised client in a global
variable, and expose its config as a class method on the `Client`
class.

### [Implement AppSignal transmitter](https://github.com/appsignal/appsignal-python/commit/762835f2ffe95c96d08f6ce197b5a9205c1845da)

Implement a transmitter module similar to the ones in other
integrations, which handles the sending of requests with the usual
AppSignal identifying params.

### [Remove empty `TYPE_CHECKING` block](https://github.com/appsignal/appsignal-python/commit/e6f695197c24cee32cd620beca85478612e504a0)

Probably a leftover from some typing refactor.

### [Implement heartbeats for Python](https://github.com/appsignal/appsignal-python/commit/e1c0e856f91b82e6f3f4444638b0ee5f63f5e790)

Implement the heartbeats functionality for Python, similarly to how
it was implemented for Ruby, Elixir and Node.js.

Due to limitations on how `with` blocks are implemented, which make
it impossible for a function to behave in different ways when called
"by itself" and when called by a `with` block, implement two different
heartbeat convenience functions, `heartbeat_with_context` and
`heartbeat`. The naming convention follows that of our existing
`send_error` and `send_error_with_context` helpers.

Add a `LOGGING_ENDPOINT` configuration variable to configure the
logging endpoint, which is also the endpoint to which heartbeats are
sent. This is a bit bizarre, as Python does not support the logging
feature yet, but it is consistent with our other integrations.


jeffkreeftmeijer (approved):

src/appsignal/heartbeat.py
```diff
@@ -0,0 +1,80 @@
...
+        logger = logging.getLogger("appsignal")
+
+        if not config.is_active():
+            logger.debug("AppSignal not active, not transmitting heartbeat event")
+            return
+
+        url = f"{config.option('logging_endpoint')}/heartbeats/json"
+        try:
+            response = transmit(url, json=event)
+            if 200 >= response.status_code >= 299:
```
tombruijn, line 47:
```suggestion
            if 200 >= response.status_code <= 299:
```
It should be lower than 300, right?

tombruijn (commented):
The build is failing because of some import error.

src/appsignal/heartbeat.py
unflxw: This is Python's somewhat strange, but intuitive once you're aware of it, chain comparison syntax! https://note.nkmk.me/en/python-chain-comparisons/

`200 >= response.status_code >= 299` means the same as `200 >= response.status_code && response.status_code >= 299`.

unflxw: Oh, wait, no, you're completely right, that is not what's implemented. It should be `200 <= response.status_code <= 299`.

unflxw:
Ended up changing the design significantly, please re-review. I feel that `heartbeat_with_context` was not a good name -- it was borrowed from `send_error_with_context`, but "with context" in that function means "with this context _for the error_", not "by using a `with` context block".

I feel that the current design, where the `Heartbeat` class implements the `with` context manager protocol, and `heartbeat` can optionally be called with a function (in a way, doing _both_ the Ruby API design and the Elixir/Node.js API design) is a lot clearer, and likely to be a better fit to most use-cases.

Also fixed some major issues with the implementation.

.changesets/implement-heartbeats.md
```diff
@@ -0,0 +1,44 @@
...
+
+You can use the `Heartbeat` class (not the `heartbeat` function!) with a
+`with` block, to report to AppSignal both when the process starts, and when
+it finishes, allowing you to see the duration of the process:
+
+```python
+from appsignal import Heartbeat
+
+def send_invoices():
+  with Heartbeat("send_invoices"):
```
tombruijn, line 25: That I can use both the capitalized and non-capitalized way is going to trip up in the future for sure. Is this a common pattern?

luismiramirez (approved):

.changesets/implement-heartbeats.md
unflxw: Thank you for pointing this out! I see the problem. It is unclear to me how to meaningfully address it. Some thoughts:

You can also use both the capitalized class and non-capitalized methods in the other integrations as well -- alongside the `heartbeat` convenience method, the `Appsignal.Heartbeat` class/module in Ruby/Elixir/Node.js is exposed for people who want to call start and finish manually, in case that's an use case that exists. 

But an important difference is that this is the only integration in which we _document_ that, because, unlike in the other integrations, one of the convenience APIs (the context manager one) is tied to the class.

I'd say it's not super uncommon to have a convenience API with simple function calls and a more flexible API based around classes. The `requests` library comes to mind: usually you just do `requests.get(...)`, but you can also create a `requests.Request(object)`. It is probably more uncommon to have them use the same name.

---

Anyway. The way to address it that comes to mind is to not document the context manager API. Document Python's function-based API, like it is done for Elixir and Node.js, and omit documenting the Ruby-style block-based API.

This saves us a lot of "that's `heartbeat` not `Heartbeat`" in the docs. We can always re-evaluate if customers reach out saying they'd like it to be exposed as a context manager.

Does this seem reasonable for now?

tombruijn: Yeah I'd be okay with simplying it a bit. Or to make it more distinct, like `Heartbeat.new()` vs `heartbeat()`

unflxw: `Heartbeat.new()` is unfortunately a Ruby-ism -- class instance initialisation in Python is always `Heartbeat()`

Amended the changelog to remove references to the uppercase-h `Heartbeat`.

PR #207, opened 2024-04-30
Add stop() to client

Juice10:
Needed to do https://docs.appsignal.com/ruby/integrations/rake.html#appsignalstop-requirement in python


Juice10:
My suggestion probably won't actually work

PR #209, opened 2024-05-03
Fix app path assertion in diagnose tests

tombruijn:
Update the submodule with the changes in
https://github.com/appsignal/diagnose_tests/pull/116

This fixes the changes in the `hatch run` behavior. It will now (correctly) use the directory in which `hatch run` is run as the working directory, instead of the project's directory.

[skip changeset]
[skip review]


tombruijn:
@ team, please update your hatch locally with `pip install hatch` so things work after pulling these changes.

tombruijn:
@appsignal/integrations please update your hatch locally with pip install hatch so things work after pulling these changes.

PR #212, opened 2024-05-29
Add noop mode for unsupported systems

### LINKED ISSUE #179: Windows "noop support"
In the other AppSignal integrations, we don't support Windows, but it is possible to install the integration on Windows, even if it doesn't do anything. With the Python integration, this is not currently possible, as we don't release a package tagged for Windows.

I'd suggest releasing a package tagged with the pip platform `py3-any-any`, which would be installed for any platform where a more specific compatible package does not exist. This package would contain no build of the agent. Our application would have to deal "well enough" with the agent not existing. (Emit error message when `appsignal.start()` is called, do nothing else)

This would also provide at least "noop support" for FreeBSD (see #49)

luismiramirez:
Add wildcard triple for unsupported systems

The `any` tag is now installed when downloading the package from an unsupported system.

Don't start components when on unsupported arch

When trying to run the agent, OpenTelemetry and probes on an unsupported system (any-any), prevent the three of them from starting.

Closes #179

### DIFF
--- a/.changesets/add-noop-mode-for-unsupported-systems.md
+++ b/.changesets/add-noop-mode-for-unsupported-systems.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Unsupported systems, like Microsoft Windows, won't start the agent and other integration components to prevent them from failing and allowing apps to be run normally.
--- a/src/appsignal/agent.py
+++ b/src/appsignal/agent.py
@@ -16,6 +16,15 @@ class Agent:
 
     def start(self, config: Config) -> None:
         config.set_private_environ()
+
+        if self.architecture_and_platform() == ["any"]:
+            print(
+                "AppSignal agent is not available for this platform. "
+                "The integration is now running in no-op mode therefore "
+                "no data will be sent to AppSignal."
+            )
+            return
+
         p = subprocess.Popen(
             [self.agent_path, "start", "--private"],
             stdout=subprocess.PIPE,
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -41,6 +41,8 @@ class Client:
         if self._config.is_active():
             logger.info("Starting AppSignal")
             agent.start(self._config)
+            if not agent.active:
+                return
             start_opentelemetry(self._config)
             self._start_probes()
         else:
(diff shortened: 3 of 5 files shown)


unflxw:
Did some thorough testing around the packaging itself, because the build hook is tricky and not covered by any automated testing and I don't trust it very much. Writing it down here in case someone wants to follow at home:

- built it with `_APPSIGNAL_BUILD_TRIPLE=any hatch build -t wheel`.
- also built it after a `hatch run build:me` (for my non-any platform) to make sure no artifacts leaked between the two builds
- used `tar -ztvf dist/appsignal-1.3.3-py3-none-any.whl` to list the files
- un-tar'd to ensure `appsignal/_appsignal_platform` was `any`.
- ran `hatch run build:all` to make sure it was built as part of the release process

All of that went fine, so the artifact is being built as expected.

src/appsignal/agent.py
```diff
@@ -16,6 +16,15 @@ class Agent:
 
     def start(self, config: Config) -> None:
         config.set_private_environ()
+
+        if self.architecture_and_platform() == ["any"]:
+            print(
+                "AppSignal agent is not available for this platform. ",
+                "The integration is now running in no-op mode therefore ",
+                "no data will be sent to AppSignal.",
```
unflxw, line 24:
```suggestion
                "AppSignal agent is not available for this platform. "
                "The integration is now running in no-op mode therefore "
                "no data will be sent to AppSignal."
```

(the commas cause it to be sent as three different arguments to `print()` -- so there's a double space after each break: the space in the string, and the space `print()` puts between arguments)

unflxw (approved):

.changesets/add-noop-mode-for-unsupported-systems.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Unsupported systems as Windows won't start the agent and other integration components to prevent them from failing and allowing apps to be run normally.
```
tombruijn, line 6:
```suggestion
Unsupported systems, like Microsoft Windows, won't start the agent and other integration components to prevent them from failing and allowing apps to be run normally.
```

tombruijn (approved):

Issue #181, opened 2023-11-30
Support Pika


Nualadoyle:
Requested by a customer on 30th November

> 1) Will Pika support (https://pika.readthedocs.io/en/stable/) happening soon for python performance insights?

- [Intercom](https://app.intercom.com/a/inbox/yzor8gyw/inbox/shared/all/conversation/16410700274820#part_id=comment-16410700274820-21909542594)
- [OpenTelemetry package for Pika](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-pika)

luismiramirez:
Pika support added in v1.3.7

- https://pypi.org/project/appsignal/1.3.7/
- https://docs.appsignal.com/python/instrumentations/pika.html

PR #217, opened 2024-06-18
Add support for Pika

luismiramirez:
Add instrumentation support for Pika messages. The global messaging extractor already covers the cases this instrumentation adds, so there's no need for a specific extractor so far.

Action with message production:

image

Consumption action automatically instrumented

image

Part of: #181

### DIFF
--- a/.changesets/add-support-for-pika.md
+++ b/.changesets/add-support-for-pika.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: add
+---
+
+Add support for Pika
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -104,6 +104,7 @@ class Config:
         "opentelemetry.instrumentation.jinja2",
         "opentelemetry.instrumentation.mysql",
         "opentelemetry.instrumentation.mysqlclient",
+        "opentelemetry.instrumentation.pika",
         "opentelemetry.instrumentation.psycopg",
         "opentelemetry.instrumentation.psycopg2",
         "opentelemetry.instrumentation.pymysql",
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -103,6 +103,12 @@ def add_mysqlclient_instrumentation() -> None:
     MySQLClientInstrumentor().instrument()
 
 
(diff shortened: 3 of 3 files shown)


luismiramirez:
Docs: https://github.com/appsignal/appsignal-docs/pull/1937

tombruijn (approved):

unflxw:
The instrumentation does some wacky stuff here that our global extractor is not ready to handle. I think it would be good to add a custom extractor in the agent to fix this.

---

There are two spans here, one for sending the span and another for receiving it, that the Pika instrumentation for Python, in the test setup, sends with the names "test send" and "test receive".

These are actually pretty great span names! They are low cardinality, and they refer to the something that is meaningful to the user (the queue name, in this case "test") and a clear action ("send", "receive")

But then, **because the attributes are not correctly set, the global extractor does not know what to do with them**, and it messes them up in different ways.

For the send span, because there is no `messaging.operation` value (it should be "publish") the global extractor mostly doesn't know what to do with it. It does not set a good category (something like "send.rabbitmq" or "publish.rabbitmq", instead of "unknown.rabbitmq") and it does not change the name (this is actually good)

For the receive span, the opposite happens. Because there is a `messaging.operation` value ("receive") the global extractor sets the category (good) and rewrites the name.

But, when it rewrites the name, because `messaging.destination` is set to the customer tag (an opaque, high-cardinality value) instead of to the queue name ("test"), it messes it up, and rewrites it as something like "receive message (ctag.68b329da9893e34099c7d8ad5cb9c940)", which is less useful for the user than the queue name.

---

Thankfully this is why we have extractors! I would suggest a very simple extractor that doesn't use the messaging global extractor at all, instead just leaving the span name as-is, and setting the category using the last word of the span name (so, "send.pika", or "receive.pika"). Does that sound reasonable?

unflxw:
Ah, the agent is also merging the trace for the consumer and the producer (because they are propagated as the same trace)

Ideally, these would be two separate traces, like in Sidekiq, Celery or BullMQ. But this is harder to fix, so maybe we can leave it like this for now.

luismiramirez:
thanks for the insight @unflxw, I'll work on an extractor

Issue #221, opened 2024-08-05
[Question] Manually Flush logging and traces 


Felipe-RA:
Greetings,

I'm using `AppSignal` to log metrics and traces for my `Serverless` Application.

While testing with  **Serverless Offline** all the lambda functions report really great on my AppSignal Dashboard.

But while my application is deployed at AWS, no error or metric shows at AppSignal (they show at CloudWatch log groups instead)

I suspect that the problem is that the AppSignal Agent does not have enough time to flush it's content to AppSignal before the lambda dies.

Can I flush all the logging manually, somehow?

This is how I imagine it:

```python
name="my-app-name"
environment="my-ENV"

appsignal = Appsignal(
    active=True,
    name=name,
    environment=environment,
    push_api_key= "MY-KEY"
)

appsignal.start()


try:

  ### ... app logic 

except:

  ### code to format the error response

finally:

  appsignal.flush() #  to manually send any opentelemetry and other traces/metrics to AppSignal
```

Is there a way to do this right now? 

Thanks for taking the time to read this!

luismiramirez:
Hi @Felipe-RA!

You can now use the `appsignal.stop()` method from AppSignal for Python v1.3.9

Felipe-RA:
Thank you @luismiramirez  !

PR #222, opened 2024-08-12
Add helper to manually stop agent processes

luismiramirez:
In some specific contexts such as serverless functions, the execution is stopped before the agent gets all the data to send it to AppSignal.

There's a new function that manually allows you to gracefully stop AppSignal's agent processes so all the data is flushed before the execution is closed.

Fixes: #221

### DIFF
--- a/.changesets/new-helper-to-manually-stop-all-agent-processes.md
+++ b/.changesets/new-helper-to-manually-stop-all-agent-processes.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+type: add
+---
+
+Add helper to manually stop the agent process for this AppSignal instance.
+
+Some contexts, like serverless functions, exit before AppSignal can ensure all data is sent to our servers. To ensure the data is sent, the new `appsignal.stop()` method can be called to gracefully stop the AppSignal agent process.
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import os
+import signal
+import time
 from typing import TYPE_CHECKING
 
 from . import internal_logger as logger
@@ -47,6 +50,19 @@ class Client:
             self._start_probes()
         else:
             logger.info("AppSignal not starting: no active config found")
+
+    def stop(self) -> None:
+        logger.info("Stopping AppSignal")
+        working_dir = self._config.option("working_directory_path") or "/tmp/appsignal"
+        lock_path = os.path.join(working_dir, "agent.lock")
+        try:
+            with open(lock_path) as file:
+                line = file.readline()
+                pid = int(line.split(";")[2])
+                os.kill(pid, signal.SIGTERM)
+                time.sleep(2)
+        except FileNotFoundError:
+            logger.info("Agent lock file not found")
 
     def _start_probes(self) -> None:
         if self._config.option("enable_minutely_probes"):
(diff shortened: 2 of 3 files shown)


unflxw (approved):
Left some comments on the current implementation, looks good.

But perhaps something to consider: instead of stopping _all_ agent processes, we could look at the `agent.lock` file in the configured working directory for the agent, which contains the PID of this agent, and stop only this one:

```bash
# from a shell inside the python/flask test setup
> cat /tmp/working_directory/agent.lock 
3286136129964684323;running;85
#                           ^^ pid!
```

Then this `appsignal.stop()` method will stop only this agent, the one that this Python application is talking to. And it also means that we don't need to manually parse the output of `ps -A`.

src/appsignal/client.py
```diff
@@ -48,6 +51,21 @@ def start(self) -> None:
         else:
             logger.info("AppSignal not starting: no active config found")
 
+    def stop_agent_processes(self) -> None:
+        logger.info("Killing all AppSignal agent processes")
+
+        processes = subprocess.Popen(["ps", "-A"], stdout=subprocess.PIPE)
```
tombruijn, line 57: This could kill agents from other apps running on the same host. It would be nice if we look up the PID in the agent's working directory for this app. In the working directory an `agent.lock` file exists that has a format like so: `config-digest;state;PID`. If you read that file, split by `;` and take the last value you have the PID for the agent started by this app.

.changesets/new-helper-to-manually-stop-all-agent-processes.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: patch
+type: add
+---
+
+New helper to manually stop all agent processes.
```
tombruijn, line 6:
```suggestion
Add helper to manually stop all agent processes.
```

.changesets/new-helper-to-manually-stop-all-agent-processes.md
```diff
@@ -0,0 +1,8 @@
...
+
+Some specific contexts as serverless functions require to ensure all the data is sent to AppSignal before finishing. To achieve this, a new method (`appsignal.stop_agent_processes()`) gracefully stops all AppSignal processes so the data is not lost when wrapping up the app or function execution.
```
tombruijn, line 8:
```suggestion
Some contexts, like serverless functions, exit before AppSignal can ensure all data is sent to our servers. To ensure the data is sent, the new `appsignal.stop()` method can be called to gracefully stop the AppSignal agent process.
```

src/appsignal/client.py
```diff
@@ -48,6 +51,21 @@ def start(self) -> None:
         else:
             logger.info("AppSignal not starting: no active config found")
 
+    def stop_agent_processes(self) -> None:
```
tombruijn, line 54:
```suggestion
    def stop(self) -> None:
```

src/appsignal/client.py
```diff
@@ -48,6 +51,19 @@ def start(self) -> None:
         else:
             logger.info("AppSignal not starting: no active config found")
 
+    def stop(self) -> None:
+        logger.info("Stopping AppSignal")
+        working_dir = self._config.option("working_dir") or "/tmp/appsignal"
```
unflxw, line 56: It seems the config option's name is [`working_directory_path`](https://github.com/appsignal/appsignal-python/blob/10422663a4723e424f7c5a663269332848f18e69/src/appsignal/config.py#L51).

unflxw (approved):

PR #227, opened 2024-10-07
Implement scheduler and heartbeat checkins

unflxw:
### [Add type hints to conftest.py](https://github.com/appsignal/appsignal-python/commit/91ca1f97122c349b0a7eab8ab53f4ecbe58da6d5)

This allows us to type-check it with mypy -- more importantly, it
makes it so tools that type-check every file in the workspace don't
freak out about every single line in it.

### [Implement scheduler and heartbeat check-ins](https://github.com/appsignal/appsignal-python/commit/48b7d039bfdf1cc7dbe35c366a3c675accce8656)

Implement a scheduler for check-in events, ensuring that the events
are sent asynchronously from the thread in which they are scheduled.

Implement a heartbeat check-in helper, which sends heartbeat check-in
events. When given the `continuous=True` flag, the events will be
sent continously from a separate thread during the duration of the
process.


luismiramirez (approved):
LGTM 🚀

PR #231, opened 2024-11-28
Lock linters to latest version

unflxw:
Lock `ruff`, `black`, `isort` and `mypy` to the latest version. We should update these from time to time.

[skip changeset]


tombruijn (approved):
Thanks!

PR #245, opened 2025-09-22
Pin `click` to earlier version

unflxw:
The `click` package, which `hatch` depends on, released a minor version with breaking changes, which are breaking our `hatch`-based builds.

Pin `click` to `8.2.x` versions to prevent this.


unflxw:
Merging this to make `main` green.

PR #249, opened 2025-10-07
Add collector attributes config options

unflxw:
This is work I realised needed doing while working on documenting the configuration options for the [collector endpoint PR](https://github.com/appsignal/appsignal-python/pull/247).

See https://github.com/appsignal/diagnose_tests/pull/142 for diagnose tests update.

---

Implement the following configuration options in the integration, mapping to the collector-supported, `appsignal.config.`-prefixed resource attribute of the same name:

- `filter_attributes`
- `filter_function_parameters`
- `filter_request_query_parameters`
- `filter_request_payload`
- `response_headers`
- `send_function_parameters`
- `send_request_query_parameters`
- `send_request_payload`

Additionally, also map the following existing configuration options in the integrations to resource attributes in the same manner:

- `filter_session_data` (mapped to `filter_request_session_data`)
- `ignore_actions`
- `ignore_errors`
- `ignore_namespaces`
- `request_headers`
- `send_session_data` (mapped to `send_request_session_data`)

When validating the configuration, if the configuration would trigger the use of the collector, emit a warning about any configuration options set to non-default values which have no effect when using the collector. Do the same when the agent would be used for configuration options that would have no effect when using the agent.

### DIFF
--- a/src/appsignal/client.py
+++ b/src/appsignal/client.py
@@ -46,6 +46,7 @@ class Client:
 
         if self._config.is_active():
             logger.info("Starting AppSignal")
+            self._config.warn()
             self._binary.start(self._config)
             if not self._binary.active:
                 return
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import os
 import platform
+import socket
 import tempfile
 from typing import Any, ClassVar, List, Literal, TypedDict, cast, get_args
 
+from . import internal_logger as logger
 from .__about__ import __version__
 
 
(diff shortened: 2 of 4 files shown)


unflxw:
As part of this change, I made it so the `hostname` configuration option is prefilled from `socket.gethostname()`. When using the agent, this is not necessary, as it will auto-detect the hostname if not provided in a similar way. However, when using the collector, the hostname is a required resource attribute.

Explaining this here because a) I forgot to mention it on the commit description and b) it explains the diagnose test failures.

unflxw:
Rebased this into #247, can be reviewed as part of it.

PR #268, opened 2026-06-08
Send request sample data in collector mode

unflxw:
Emit request parameters and headers under the attribute names the
collector and server recognize when running in collector mode. Parameters
move from `appsignal.request.parameters` to `appsignal.request.payload`,
and headers from the `appsignal.request.headers` prefix to the
`http.request.header` semantic-convention prefix.

Neither the collector nor the server recognized the previous names, so
this sample data was silently dropped from the trace UI in collector
mode. Agent mode keeps the existing names. The Django and Flask hooks now
route through `set_params` so they pick up the branch too.

### DIFF
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -58,33 +58,29 @@ def add_celery_instrumentation(_config: Config) -> None:
 
 
 def add_django_instrumentation(_config: Config) -> None:
-    import json
-
     from django.http.request import HttpRequest
     from django.http.response import HttpResponse
     from opentelemetry.instrumentation.django import DjangoInstrumentor
 
+    from .tracing import set_params
+
     def response_hook(span: Span, request: HttpRequest, response: HttpResponse) -> None:
-        span.set_attribute(
-            "appsignal.request.parameters",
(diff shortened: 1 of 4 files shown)


src/appsignal/opentelemetry.py
```diff
@@ -58,33 +58,29 @@ def add_celery_instrumentation(_config: Config) -> None:
...
     from opentelemetry.instrumentation.django import DjangoInstrumentor
 
+    from .tracing import set_params
+
     def response_hook(span: Span, request: HttpRequest, response: HttpResponse) -> None:
-        span.set_attribute(
-            "appsignal.request.parameters",
-            json.dumps({"GET": request.GET, "POST": request.POST}),
-        )
+        set_params({"GET": request.GET, "POST": request.POST}, span)
```
tombruijn, line 68: When we eventually switch this over to report to the collector only, I would want this to set the GET values on the `appsignal.request.query_parameters` attribute and the payload on the `appsignal.request.payload` attribute.
Now it puts everything on the payload. That's fine for now while we support both reporting methods.

tombruijn (approved):

PR #270, opened 2026-08-21
Use `app` as the default service name

unflxw:
### [Use `app` as the default service name](https://github.com/appsignal/appsignal-python/commit/fcf33d89d916aedd1fcf8488599f13df2e8df2f8)

The collector turns the service name into names that customers see. A
trace's namespace becomes `{service name}/{namespace}`, so a web
request from an app that does not set `service_name` was filed under
`unknown/web`. A log line's group falls back to the service name on
its own, so those lines were grouped under `unknown`.

Leaving `service_name` unset is the normal case for an app that is a
single service. Nothing about it is actually unknown, so the fallback
now describes what it is. The `revision` and `hostname` fallbacks
still report `unknown`, because those are facts about the running
process that AppSignal really does not have.

### [Apply revision and host name fallbacks to None](https://github.com/appsignal/appsignal-python/commit/3e289ce679a7844a5b3b0dda0abd14bddbb05a9f)

The OpenTelemetry resource looks up `revision` and `hostname` with a
dictionary default, which only applies when the key is missing. Passing
either option as `None` puts the key in the config with a `None` value,
so the lookup returns that `None` and the resource builder drops the
attribute. The collector supplies its own revision when the attribute is
absent, but not its own host name, so an app configured with
`hostname=None` reported no host name at all.

### DIFF
--- a/.changesets/default-service-name-to-app.md
+++ b/.changesets/default-service-name-to-app.md
@@ -1,0 +1,7 @@
+---
+bump: patch
+type: change
+---
+
+Report `app` instead of `unknown` as the OpenTelemetry service name when the
+`service_name` configuration option is not set and collector mode is in use.
--- a/.changesets/report-host-name-when-set-to-none.md
+++ b/.changesets/report-host-name-when-set-to-none.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+type: fix
+---
+
+Report `unknown` as the host name in collector mode when the `hostname`
+configuration option is set to `None`. Apps that set it that way reported no
+host name at all.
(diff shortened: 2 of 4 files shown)


unflxw:
Rationale for this (and for consistency, its twin Ruby PR at https://github.com/appsignal/appsignal-ruby/pull/1581) is to have the same name for customers who upgrade from agent mode to collector mode. @matsimitsu will be changing the processor's translation layer to use "app" as well, so we have a shared, reasonable enough default.

Docs will still advise setting a service name as part of collector mode setup, so the fallback may not be very relevant, but customers who upgrade somewhat manually may skip the docs.

PR #276, opened 2026-08-25
Send app_path and ignore_logs to the collector

### LINKED ISSUE #273: Send app_path as a resource attribute in collector mode
The collector understands `appsignal.config.app_path`, and the package never sends it. The value already exists: `Config.load_from_system` sets `app_path` to the working directory, and `set_private_environ` passes it to the agent as `_APPSIGNAL_APP_PATH`. Only `_resource` in `src/appsignal/opentelemetry.py` leaves it out.

Sending it is all that is needed. Nothing has to change in the collector or the processor.

The collector reads `appsignal.config.app_path` and, unlike the other AppSignal config attributes, keeps it on the outgoing span batch instead of stripping it. See `KEEP_APPSIGNAL_RESOURCE_CONFIG` in `collector/src/tracing/traces_buffer.rs`. The processor reads it back off the span's resource attributes and uses it to build the backtrace parser.

The app path is what gets stripped from each backtrace line. Without it, the lines keep their absolute paths. It gets worse from there, because the processor decides whether a frame belongs to the application by checking whether its path is relative. Every frame now starts with a `/`, so none of the application's own frames are recognised as app lines.

The Ruby gem has the same gap, tracked in https://github.com/appsignal/appsignal-ruby/issues/1585. That issue also covers `ignore_logs`, which this package has no option for yet. Adding one is tracked separately in #275.

### LINKED ISSUE #275: Add the ignore_logs option now that collector mode sends logs
The package has no `ignore_logs` option. Other integrations have had one for a while, and the Ruby gem's version is documented as `ignore_logs` with the `APPSIGNAL_IGNORE_LOGS` environment variable and an array of strings as its value.

The reason the package never had one was that there were no logs to filter. `Config.should_instrument_logging` says so directly: "The agent does not support logging, so this is equivalent to whether it should use a collector."

Collector mode changes that. The package now attaches an OpenTelemetry `LoggingHandler` to the root logger and sends log records through the collector, so there are logs, and the collector is the one component that can filter them. It reads the `appsignal.config.ignore_logs` resource attribute and applies the same matching the AppSignal extension applies in agent mode for other languages, including the small subset of `^`, `$` and `.*` that the option supports. See `collector/src/logs/logs_buffer.rs`.

So the work is to add the option, and then send it as a resource attribute alongside the other `appsignal.config.*` attributes in `_resource`. Nothing has to change in the collector.

One thing to decide while adding it: in the other integrations `ignore_logs` applies in both modes, but here it can only ever apply in collector mode, because agent mode sends no logs at all. That is worth saying in the option's documentation so nobody sets it in agent mode and expects it to do something.

The equivalent gap in the Ruby gem is different, because the gem already has the option and simply does not send it. That is tracked in https://github.com/appsignal/appsignal-ruby/issues/1585, together with `app_path`.

unflxw:
Fixes https://github.com/appsignal/appsignal-python/issues/273.
Fixes https://github.com/appsignal/appsignal-python/issues/275.

### [Send the app path to the collector](https://github.com/appsignal/appsignal-python/commit/bb3759a2a0a015c44a31b37718d33396d948367e)

The processor strips the app path from each backtrace line, and decides
whether a frame belongs to the application by checking that its path is
relative. Without the app path, every frame keeps its absolute path, so
no line is recognized as the application's own.

### [Add the ignore_logs option](https://github.com/appsignal/appsignal-python/commit/723fc30bb5fde2899562028c20c937b6bce791b2)

Collector mode sends log records, and the collector filters out the ones
matching this option when the resource carries it. The package had no
option for it because agent mode sends no logs at all, which is also why
it is not passed on to the agent.

The option is documented in https://github.com/appsignal/appsignal-docs/pull/192,
including that it only has an effect in collector mode.

### DIFF
--- a/.changesets/add-the-ignore-logs-option.md
+++ b/.changesets/add-the-ignore-logs-option.md
@@ -1,0 +1,8 @@
+---
+bump: minor
+type: add
+---
+
+Add an `ignore_logs` option, which takes a list of patterns. Log lines that match any of the patterns are not sent to AppSignal. Set it with the `APPSIGNAL_IGNORE_LOGS` environment variable, or as an option on the `Appsignal` client. Read our [ignore logs guide](https://docs.appsignal.com/guides/filter-data/ignore-logs.html) for the patterns that are supported.
+
+This option only has an effect in collector mode, because that is the only mode in which this package sends logs.
--- a/.changesets/send-the-app-path-to-the-collector.md
+++ b/.changesets/send-the-app-path-to-the-collector.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+In collector mode, backtrace lines from your own application are now shown as paths relative to your application's root, and are recognized as your application's code.
--- a/src/appsignal/config.py
+++ b/src/appsignal/config.py
@@ -40,6 +40,7 @@ class Options(TypedDict, total=False):
     http_proxy: str | None
     ignore_actions: list[str] | None
     ignore_errors: list[str] | None
+    ignore_logs: list[str] | None
     ignore_namespaces: list[str] | None
     log: str | None
     log_level: str | None
@@ -224,6 +225,7 @@ class Config:
             http_proxy=os.environ.get("APPSIGNAL_HTTP_PROXY"),
             ignore_actions=parse_list(os.environ.get("APPSIGNAL_IGNORE_ACTIONS")),
             ignore_errors=parse_list(os.environ.get("APPSIGNAL_IGNORE_ERRORS")),
+            ignore_logs=parse_list(os.environ.get("APPSIGNAL_IGNORE_LOGS")),
             ignore_namespaces=parse_list(os.environ.get("APPSIGNAL_IGNORE_NAMESPACES")),
             log=os.environ.get("APPSIGNAL_LOG"),
             log_level=os.environ.get("APPSIGNAL_LOG_LEVEL"),
@@ -433,6 +435,7 @@ class Config:
             "filter_function_parameters",
             "filter_request_payload",
             "filter_request_query_parameters",
+            "ignore_logs",
             "response_headers",
             "send_function_parameters",
             "send_request_payload",
--- a/src/appsignal/opentelemetry.py
+++ b/src/appsignal/opentelemetry.py
@@ -311,6 +311,7 @@ def _resource(config: Config) -> Resource:
             ),
             "appsignal.config.ignore_actions": config.options.get("ignore_actions"),
             "appsignal.config.ignore_errors": config.options.get("ignore_errors"),
+            "appsignal.config.ignore_logs": config.options.get("ignore_logs"),
             "appsignal.config.ignore_namespaces": config.options.get(
                 "ignore_namespaces"
             ),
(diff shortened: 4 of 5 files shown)


.changesets/add-the-ignore-logs-option.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: minor
+type: add
+---
+
+Add an `ignore_logs` option, which takes a list of patterns. Log lines that match any of the patterns are not sent to AppSignal. Set it with the `APPSIGNAL_IGNORE_LOGS` environment variable, or as an option on the `Appsignal` client.
```
unflxw: This could link to the page that documents how to ignore logs, which has examples of the patterns that are allowed.

.changesets/send-the-app-path-to-the-collector.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+In collector mode, backtrace lines from your own application are now shown as paths relative to your application's root, and are recognized as your application's code. Before, they were shown as absolute paths, and no line was marked as coming from your application.
```
unflxw:
```suggestion
In collector mode, backtrace lines from your own application are now shown as paths relative to your application's root, and are recognized as your application's code.
```

.changesets/add-the-ignore-logs-option.md
unflxw: Added a link to the ignore logs guide.

.changesets/send-the-app-path-to-the-collector.md
unflxw: Applied.

src/appsignal/config.py
```diff
@@ -224,6 +225,7 @@ def load_from_environment() -> Options:
             http_proxy=os.environ.get("APPSIGNAL_HTTP_PROXY"),
             ignore_actions=parse_list(os.environ.get("APPSIGNAL_IGNORE_ACTIONS")),
             ignore_errors=parse_list(os.environ.get("APPSIGNAL_IGNORE_ERRORS")),
+            ignore_logs=parse_list(os.environ.get("APPSIGNAL_IGNORE_LOGS")),
```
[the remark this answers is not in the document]
unflxw: Dismissed. `parse_list` is shared by every list option in this package, and the AppSignal integrations and the collector all split these lists on a comma without trimming. Trimming here would make `ignore_logs` behave differently from every other list option, and differently from the same option in the other integrations. Worth raising on its own if we want it everywhere.

.changesets/add-the-ignore-logs-option.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: minor
+type: add
+---
+
+Add an `ignore_logs` option, which takes a list of patterns. Log lines that match any of the patterns are not sent to AppSignal. Set it with the `APPSIGNAL_IGNORE_LOGS` environment variable, or as an option on the `Appsignal` client. Read our [ignore logs guide](https://docs.appsignal.com/guides/filter-data/ignore-logs.html) for the patterns that are supported.
```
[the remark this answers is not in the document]
unflxw: Dismissed. This is the wording our documentation already uses for `ignore_logs` in every integration: "Any log message containing any of the elements of the list will not be transmitted to AppSignal." Which hop drops the line is an internal detail, and a changelog entry that describes it differently from the option documentation would be more confusing than helpful.

lipskis (approved):

